PR #32, opened 2016-05-14
another few clippy warnings fixed

llogiq:
Using the `languages.values()` iterator may save some time/space, as does using a character as pattern instead of a one-character string.


Aaronepower:
This PR request is based on the old version, so the fixes are no longer valid.

llogiq:
Ouch – my git client failed me. I'll pull from a clean repo again and re-try :smile:

PR #35, opened 2016-05-17
Prints code lines before comments and blanks

liigo:
Most users are more concerned about code lines than comments or blanks.


Aaronepower:
Thanks!

PR #40, opened 2016-05-26
Fixed parameter count mismatch.

Phlosioneer:
Tests weren't updated to include the correct number of parameters when has_trailing_comments was updated (c5750f9).
nested=true uses the original, pre-change codepath for the tests.


Aaronepower:
Ah, yeah should of done that. Thanks!

PR #49, opened 2016-07-12
Document that `--exclude` accepts a list of values

acdha:
I'm not entirely happy with the wording but wanted to avoid using terms like “match” to avoid giving the impression that things like regular expressions or shell globs were supported.


Aaronepower:
@acdha Maybe if it would be better if `containing a term` was changed `containing any of the strings`?

PR #59, opened 2016-09-10
Made a series of modifications to the languages comments:

mwilli20:
Clojure - Removed #
	Forth   - ( Comment) style comments need a space after the opening paren
	Haskell - Added multiline comment style {- Block comment -}
	Jai     - Has nested block comments
	Julia   - Has nested block comments
	Kotlin  - Has nested block comments
	Pascal  - Pascal should be multiline from { or (* to } or *)
	Perl    - Perl5 and earlier for multiline comments need =pod to =cut.
				Perl6 will be different but as it is not finished I'm not going
				to add it.
	Python  - Python can use """ or ''' for comments
	Swift   - Has nested block comments
The is_fortran code was editted by rustfmt.

### DIFF
--- a/src/lib/language/languages.rs
+++ b/src/lib/language/languages.rs
@@ -72,11 +72,7 @@ fn count_files(language_tuple: &mut (&LanguageType, &mut Language)) {
             // FORTRAN has a rule where it only counts as a comment if it's the first
             // character in the column, so removing starting whitespace could cause a
             // miscount.
-            let line = if is_fortran {
-                line
-            } else {
-                line.trim_left()
-            };
+            let line = if is_fortran { line } else { line.trim_left() };
 
             if line.trim().is_empty() {
                 stats.blanks += 1;
@@ -265,7 +261,7 @@ impl Languages {
             Batch => Language::new_single(vec!["REM"]),
             C => Language::new_c(),
             CHeader => Language::new_c(),
-            Clojure => Language::new_single(vec![";","#"]),
+            Clojure => Language::new_single(vec![";"]),
             CoffeeScript => Language::new(vec!["#"], vec![("###", "###")])
                                      .set_quotes(vec![("\"", "\""), ("'", "'")]),
             ColdFusion => Language::new_multi(vec![("<!---", "--->")]),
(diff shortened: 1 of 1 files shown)


mwilli20:
@Aaronepower  Both Idris and Haskell are supposed to have the same comment style, so I updated the two of them to be nested. Good catch.

Aaronepower:
@mwilli20 Looks good, can you provide a resources showing JAI, Kotlin, Swift, and Julia have nested comments?

mwilli20:
@Aaronepower 
Jai - http://jai-lang.wikia.com/wiki/JAI_Lang_Wikia#Comments
Kotlin - https://kotlinlang.org/docs/reference/basic-syntax.html - under the comments section
Swift - https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/TheBasics.html - under the comments section
Julia - http://docs.julialang.org/en/release-0.4/stdlib/punctuation/ - in parenthesis

mwilli20:
@Aaronepower Also to bring these odd scenarios. Batch :: or just : is a function that does nothing. So a lot of people use that for comments. I don't know if that really counts or not. Common lisp has a (comment ) function which compiles the code but drops it, kind of like rust code in documentation test comments. Not sure if that counts either. I thought I would just let you know.

Aaronepower:
@mwilli20 Can you show examples of these? Tokei would support these if they 100% can only be used as comments. Whether they are thrown away by the compiler or included but skipped is irrelevant. Tokei works on how humans understand the code.

mwilli20:
@Aaronepower  This has a good example of using :: for comments in bash:
 http://www.tutorialspoint.com/batch_script/batch_script_comments.htm

It was clojure that added it not common lisp. https://clojuredocs.org/clojure.core/comment
This would be tricker than the Batch as you would need to go from 
( comment to the matching closing paren ).

mwilli20:
@Aaronepower But it looks like the case of clojure since it technically returns a value, this could be used in code if you wanted to.

Aaronepower:
@mwilli20 Yeah, I don't think the clojure one should be added due to both it's difficulty to implement, and that it returns a value. The Batch one though is a single line comment.

mwilli20:
@Aaronepower I can add it in on the next pull request for the strings if you want.

PR #60, opened 2016-09-10
Closes Issue 52

mwilli20:
All the changes are in the commit. 

Two small issues to be brought up are Php has a similar new comment type like the clojure one that I assume you wouldn't handle.

The other is Python doc strings are usually in """ """ or ''' ''' but this is also a multline string.


Aaronepower:
@mwilli20 Can you show me this php thing? Also can you rebase this on top of the master?

Issue #52, opened 2016-08-21
Update existing languages with their String litreals.


Aaronepower:
Now with the inclusion handling comments while in quotes, and out of them. All existing languages need to be updated with their string type as currently the default is `(", ")` which obviously doesn't apply to all languages. All that is needed to go to `src/lib/language/languages.rs` and use the `set_quotes()` method.

- [ ] ActionScript
- [ ] Ada
- [x] Assembly
- [ ] Autoconf
- [x] BASH
- [x] Batch
- [x] C
- [x] C Header
- [x] Clojure
- [x] CoffeeScript
- [x] ColdFusion
- [x] ColdFusion CFScript
- [ ] Coq
- [x] C++
- [x] C++ Header
- [ ] C#
- [ ] C Shell
- [x] CSS
- [ ] D
- [x] Dart
- [ ] Device Tree
- [ ] Erlang
- [ ] Forth
- [ ] FORTRAN Legacy
- [ ] FORTRAN Modern
- [ ] Go
- [ ] Handlebars
- [x] Haskell
- [x] HTML
- [x] Idris
- [ ] Isabelle
- [x] JAI
- [x] Java
- [x] JavaScript
- [x] Julia
- [x] JSON
- [ ] JSX
- [x] Kotlin
- [ ] LESS
- [ ] LD Script
- [ ] LISP
- [ ] Lua
- [ ] Makefile
- [x] Markdown
- [ ] Mustache
- [ ] Nim
- [ ] Objective C
- [ ] Objective C++
- [ ] OCaml
- [ ] Oz
- [ ] Pascal
- [ ] Perl
- [ ] Polly
- [x] PHP
- [ ] Protocol Buffers
- [ ] Prolog
- [ ] Python
- [ ] QCL
- [ ] R
- [ ] Ruby
- [ ] Ruby HTML
- [x] Rust
- [ ] Sass
- [ ] Scala
- [ ] Standard ML
- [ ] SQL
- [ ] Swift
- [ ] TeX
- [x] Plain Text
- [ ] TOML
- [ ] TypeScript
- [ ] Vim Script
- [ ] Unreal Script
- [ ] Wolfram
- [ ] XML
- [ ] YAML
- [x] Zsh

abhiQmar:
Can you please give an example? I don't quite get it.

```
CoffeeScript => Language::new(vec!["#"], vec![("###", "###")])
                                     .set_quotes(vec![("\"", "\""), ("'", "'")])
```
[...]

PR #61, opened 2016-09-10
Fixes issue #52

mwilli20:
Sorry about the last branch, I forgot I made the in github edit, and couldn't figure out what was wrong.
In response to the Php strings, they have a new string type in Php 5.

<<<EOD
Example of string
spanning multiple lines
using heredoc syntax.
EOD;

<<< Starts the comment, and then gets a variable name, and doesn't end until that variable name comes up again.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -8,5 +8,6 @@ The list of languages is sorted alphabetical.
 * Chris Palmer (https://github.com/PhnxRbrn)
 * Aaron Power (https://github.com/Aaronepower)
 * Luthaf (https://github.com/Luthaf)
+* Matthew Williamson (https://github.com/mwilli20)
 
 ## Translations
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Aaronepower <theaaronepower@gmail.com>"]
 build = "src/lib/build.rs"
-description = "Count code, quickly."
+description = "A program that allows you to count code, quickly."
 homepage = "https://aaronepower.github.io/tokei/"
 include = ["src/**/*", "cli.yml", "Cargo.toml", "LICENCE-APACHE", "LICENCE-MIT"]
 license = "MIT/Apache-2.0"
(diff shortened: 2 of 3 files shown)


Aaronepower:
@mwilli20 I think tokei should implement those style of comments, however that is out of scope for this issue. It MR looks fine, however can you either post here, or amend the commit message to say what exactly has been done? Just so the commit history is nicer, and I can use it later when I want to update the changelog.

mwilli20:
Modification on comments:
	Batch  - Add ::
	Lisp   - Comments can be nested (www.rosettacode.org/wiki/Comments#Common_Lisp)
	Python - Reverted to hash, based on comment thread from pull #59 as
		string has a return value

Changes to the quotes
	Assembly     -  Uses ' ' or " "
	Bash              - Uses ' ' or " "
	Batch             -  They don't use quotes for strings
	Cold Fusion - Uses ' ' or " "
	D                      - Uses " " or ` `
	Dart                - Uses " " or ' ' or """  """ or ''' '''
	Forth             - Uses " " but new, doesn't have a preset
	Fortrans       - Use " " or ' '
	Idris                - Uses " " or """ """
	Julia               - Uses " " or """ """
	Kotlin             - Uses " " or """ """
	Moustache  - Uses " " or ' '
	Nim                 - Uses " " or """ """
	Pascal           - Uses ' '
	Perl                 - Uses " " or ' '
	Php                 - Uses " " or ' '
	Python          - Uses " " or ' ' or """  """ or ''' '''
	Ruby              - Uses " " or ' '
	Sass               - Uses " " or ' '
	Sql                   - Uses ' '
	Toml               - Uses " " or ' ' or """ """ or ''' '''
	Typescript   - Uses " " or ' ' or ` `
	Vimscript     - Uses " " or ' '
	Yaml               - Uses " " or ' '
	Zsh                 - Uses " " or ' '

mwilli20:
@Aaronepower Do you want to make an issue for Php or should I make one?

Aaronepower:
@mwilli20 I'll do it you've done enough. :smile:

mwilli20:
@Aaronepower Hope it helps

PR #71, opened 2016-10-13
Add support for elm

not-fl3:
This PR adds support for Elm language. 

I just copy-pasted haskell's comment style, it's right for Elm's syntax, hope it's right from tokei's point.


Aaronepower:
@not-fl3 Thanks for the pull request! This PR looks fine, and if you could reorder it to be in alphabetical order, that would be great!

not-fl3:
Oh, my bad. Reordered.

PR #72, opened 2016-10-15
Add GLSL support

svenstaro:
I followed the [official extensions](https://github.com/KhronosGroup/glslang#execution-of-standalone-wrapper) for this.


Aaronepower:
@svenstaro Tjanks for the pull request! There are just some minor changes that need to be made before I can merge this in. 

Remove GLSL from the CHANGELOG that only gets updated when there is a new Cargo version. I plan on releasing one soon and will include it with that release.

 The enum name should be changed from `GLSL` to `Glsl` as that is the preferred style in rust code. The string version should remain the same. Thanks!

svenstaro:
@Aaronepower addressed!

PR #75, opened 2016-11-15
Only add regular files to path list

rmbreak:
This fixes multiple issues. Before this commit, a directory with
a known file extension in its name would be added to the list of
"files" to open and read from. Another issue occurred with broken
symbolic links being added to the files list.

### DIFF
--- a/src/lib/utils/fs.rs
+++ b/src/lib/utils/fs.rs
@@ -102,3 +102,27 @@ pub fn get_extension<P: AsRef<Path>>(path: P) -> Option<String> {
     }
 
 }
+
+
+#[cfg(test)]
+mod test {
+    extern crate tempdir;
+    use super::*;
+    use std::fs::create_dir;
+    use language::languages::Languages;
+    use language::LanguageType;
+    use self::tempdir::TempDir;
+
+
+    #[test]
+    fn walker_directory_as_file() {
+        let tmp_dir = TempDir::new("test").expect("Couldn't create temp dir");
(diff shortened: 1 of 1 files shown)


Aaronepower:
@rmbreak Thanks for the PR! I was wondering if you could also add a test case to prevent future breakage. tempdir is already a test dependency. If you need help with it, feel free to ask me.

rmbreak:
Just wrote test for ensuring directories are not added to list of Language files. The symbolic link test is a little more involved I think. Symbolic links are unix specific (https://doc.rust-lang.org/std/os/unix/fs/fn.symlink.html), so I figure there needs to be a unix-only test that gets run for this?

What do you think @Aaronepower?

Aaronepower:
@rmbreak Well as that link says you can create symlinks on Windows, however it requires special permissions, which would prevent the tests from passing on a regular users account so this is fine by me.

src/lib/utils/fs.rs
```diff
@@ -48,7 +48,9 @@ pub fn get_all_files<'a>(paths: Cow<'a, [&'a str]>,
                         get_language!(languages, &path)
                     };
 
-                    language.files.push(path.to_owned());
+                    if path.metadata().unwrap().is_file() {
```
Aaronepower: Can you change this, and the entry one from `unwrap` to `rs_error!`.

rmbreak: Made the changes.

Aaronepower:
Thanks!

PR #77, opened 2016-11-26
Exit gracefully instead of panicking when using a feature that isn't compiled in

Michael-F-Bryan:
At the moment if you try to output in a particular format and forgot to compile with that feature, the program will just panic.

```
$ tokei src --output json
thread 'main' panicked at 'This version of tokei was compiled without any serialization
    formats, to enable serialization, reinstall tokei with the features flag.

        cargo install tokei --features all
', /home/michael/.cargo/registry/src/github.com-1ecc6299db9ec823/tokei-4.4.0/src/main.rs:292
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Although the error message tells you exactly what the problem is and how to fix it, it would still be a lot nicer in terms of end user experience to print the error message to stderr and then exit. Normally when I see that a program has panicked it's because of a bug or an error which can't be recovered from, whereas this is just a case of you not having included a certain feature when installing.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,6 +4,7 @@
 
 #[macro_use]
 extern crate clap;
+#[macro_use]
 extern crate log;
 extern crate env_logger;
 // #[cfg(feature = "cbor")]
@@ -22,6 +23,7 @@ use std::borrow::Cow;
 #[cfg(feature = "io")]
 use std::collections::BTreeMap;
 use std::thread;
+use std::error::Error;
 use std::time::Duration;
 use std::sync::mpsc::channel;
 use std::io::{Write, stderr};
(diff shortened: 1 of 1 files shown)


Aaronepower:
Thanks for the PR, this is a super nit picky detail, if you could remove the empty line on line 254 I'd be happy to merge it in.

Michael-F-Bryan:
Done, thanks for spotting that.

Aaronepower:
@Michael-F-Bryan Sorry, I updated that comment with extra feedback!

Michael-F-Bryan:
@Aaronepower, how's that look? I also had to throw in a `#[macro_use]` line so you get access to `log`'s `error!()` macro.

Aaronepower:
Yeah, that's fine.

PR #79, opened 2016-11-27
Drop tx before reading from rx to prevent blocking forever

mbrubeck:


mbrubeck:
Fixed by f732e7cb0d1e4b0a80c82a0be178ab3956cc3b56

PR #81, opened 2016-11-29
Add GDScript support

GungnirInd:
Adds support for [Godot Engine](https://godotengine.org)'s GDScript scripting language.  File extension is `.gd`; comments are marked with `#` like Python.

I wasn't sure what the `quotes` section did, so I omitted it (in my very quick testing, it didn't seem to make a difference), but I can add it if needed.


Aaronepower:
quotes are the term I used to define what makes up a String literal. So for example in JavaScript `let x = "Hello World"` `(", ")` are the start and end of the string literal. If you omit it, it will default to `[(", ")]`. This is needed in edge cases as shown below.

GungnirInd:
Thanks for the comments!  I'll make changes tonight once I can get back to a computer.

Aaronepower:
Sorry, I didn't actually explain the edge cases. But below shows it clearly, without string literal detection this would be intepreted as 2 lines of code, and one comment. When in reality it is three lines of code. PS: Your code will fail the build, don't worry about it, it's unrelated.
```javascript.
let message = "Hello 
//
World"
```

GungnirInd:
Updated with name change and quotes

Aaronepower:
Thanks, I've merged it in.

PR #84, opened 2016-12-04
Differentiate between bash and sh

tuncer:
It is more accurate to not consider bash and sh to be the same.  Bash is
a superset of sh. Same goes for tcsh, zsh, fish, rc, es, etc.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -78,6 +78,26 @@
                 "in"
             ]
         },
+        "Sh":{
+            "name":"sh",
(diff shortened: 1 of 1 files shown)


Aaronepower:
Thanks for the PR, you need to also remove `"sh"` extension from bash.

tuncer:
Yeah, I've accidentally pushed the wrong commit. Fixed.

PR #87, opened 2016-12-22
Add F* support

tuncer:


tuncer:
I based this on the definition for OCaml, and it seems to share the same fault of not identifying comments. I'm not sure where the definitions for `"base":"func"` are inherited from.

tuncer:
Configuring comment syntax in `"multi":[]` and removing `"base":"func"` doesn't make a difference.

Aaronepower:
Thanks for the PR. Actually having base does make a difference as it allows tokei to re use the FUNC_REGEX, over creating a new REGEX. If you could switch it back to `"base": "func"`, I'd be happy to merge it in.

tuncer:
> If you could switch it back to "base": "func", I'd be happy to merge it in.

Done.

tuncer:
So why are comments not listed in the results for OCaml, Coq, F*, etc?

tuncer:
> Actually having base does make a difference

Code reuse is why I had it in the first version, but since the explicit multi comment definition is the same as in base:func, there's no point. It made no difference in the sense that comments are still uncounted.

Why are comments not listed in the results for OCaml, Coq, F*, etc?

Aaronepower:
I'm not sure what you mean, can you provide a sample code base where the comments aren't being counted?

tuncer:
Let's discuss it in #89.

Aaronepower:
I'm going to add merge this now, as this seems like the correct implementation, the problem with FUNC_REGEX is a seperate issue.

PR #94, opened 2017-01-15
Extend Standard ML name

tuncer:
Since Standard ML is equally often if not more referred to as SML, it
makes sense to include "(SML)" in the display name.


Aaronepower:
Can you rebase this on the latest master? Thank you.

tuncer:
Done.

PR #96, opened 2017-01-18
fixed `--languages` option always conflicting with input

m1el:
As defined in `cli.yml`, `--languages` conflicts with `<input>`,
but if `<input>` has default value, `--languages` will always conflict,
even if no inputs were specified.

Computation of the default value for `<input>` is now moved to `src/main.rs`,
because it cannot be specified in `cli.yml` along with `--languages`.

I suspect this bug was introduced in commit b7a892f4753d, when default
value for `<input>` was added.

### DIFF
--- a/cli.yml
+++ b/cli.yml
@@ -29,7 +29,6 @@ args:
         index: 1
         multiple: true
         required: true
-        default_value: "."
     - languages:
         conflicts_with:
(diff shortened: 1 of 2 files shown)


Aaronepower:
Thanks for the PR!

PR #101, opened 2017-01-29
Added support for PureScript

svisser:
This adds support for PureScript (http://purescript.org).

https://github.com/purescript/documentation/blob/master/language/Syntax.md#comments


Aaronepower:
Thanks for your PR!

PR #103, opened 2017-02-08
Update `env_logger` to 0.4

tjodden:
This avoids bringing in another version of `regex` (and it's dependencies).


Aaronepower:
Thanks, and congats on your first PR!

PR #109, opened 2017-02-23
Support "filenames" key for Makefile-style languages

kazimuth:
Also add support for Dockerfiles, Rakefiles, and CMake.

I also revised the test harness to check for no-or-multiple-language detection (because it was passing the tests when no dockerfile was detected).

The `filenames` key overrides the `extensions` key. It allows detecting `CMakeLists.txt` as `CMake` rather than `Text`. It also overrides shebangs, which may or may not be the right thing?


build.rs
```diff
@@ -62,26 +62,30 @@ fn generate_tests(out_dir: &OsStr) {
...
+        let name = path.file_stem().unwrap().to_str().unwrap().to_lowercase();
 
         string.push_str(&format!(r#"
         #[test]
         fn {0}() {{
             let mut languages = Languages::new();
             languages.get_statistics(vec!["{1}"], Vec::new());
+
+            let languages = languages.into_iter().collect::<Vec<_>>();
+            assert_eq!(languages.len(), 1, "wrong number of languages found");
```
Aaronepower: This error message should be clearer. Something like `Expected just cpp, found: ...`

CONTRIBUTING.md
```diff
@@ -89,6 +89,43 @@ syntax.
...
+Note that filenames will *override* extensions, so with the following
+configuration:
+
+```json
+"Toml":{
+    "extensions": [
+        "toml"
+    ]
+},
+"Cargo":{
```
Aaronepower: This a bad example, as if someone made a PR with Cargo I would reject it. Consider an example using `Rakefile`, or `Makefile

src/language/language_type.hbs.rs
```diff
@@ -79,6 +99,8 @@ impl LanguageType {
             None
         }
     }
+
+
```
Aaronepower: These empty lines should be removed.

Aaronepower:
Thanks for the PR! I've done a review of the code and made some comments of things I feel need to be improved before it can be merged in.

build.rs
```diff
@@ -62,26 +62,30 @@ fn generate_tests(out_dir: &OsStr) {
...
-        let name = path.file_stem().unwrap().to_str().unwrap();
+        let name = path.file_stem().unwrap().to_str().unwrap().to_lowercase();
 
         string.push_str(&format!(r#"
         #[test]
         fn {0}() {{
             let mut languages = Languages::new();
             languages.get_statistics(vec!["{1}"], Vec::new());
+
+            let languages = languages.into_iter().collect::<Vec<_>>();
```
Aaronepower: This is not needed. `languages` already has both `.len` and `.into_iter`

Aaronepower:
Closing due to inactivity. Feel free to reopen with the requested changes though!

PR #110, opened 2017-02-23
Add support for the fish shell

seiks:
Greetings from Finland! 👋

I've been a long time fan of tokei but unfortunately it doesn't recognize my fish scripts. This PR adds a language definition for the fish shell.

If I understood correctly, you don't have to do much else to support a language. Please let me know if there's any issues with this PR.

Cheers!

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -124,6 +124,25 @@
                 "bash"
             ]
         },
+        "Fish":{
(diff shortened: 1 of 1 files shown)


Aaronepower:
Hello, thanks for the PR!

Issue #111, opened 2017-02-26
Question: How can i exclude many folders with --exclude?


gabomgp:
I'm trying count code in a folder, but i want to exclude some subfolders. I think --exclude only permit exclude one subfolder. How can i exclude many folders with --exclude?

Aaronepower:
This has now been resolved.

PR #114, opened 2017-03-08
Allow multiple --exclude options

m1el:
There is no reason *not* to allow multiple exclude values.
Multiple exclude options are processed correctly by [src/main.rs#L54-L55](https://github.com/m1el/tokei/blob/c0cb7427b440307f0ffd73d51992042ac3d392d1/src/main.rs#L54-L55), but clap config doesn't accept multiple exclude options.
This pull request should cover #111.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -34,7 +34,7 @@ fn main() {
         (version: crate_version!())
         (author: "Aaron P. <theaaronepower@gmail.com> + Contributors")
         (about: crate_description!())
-        (@arg exclude: -e --exclude +takes_value "Ignore all files & directories containing the word.")
+        (@arg exclude: -e --exclude +takes_value ... "Ignore all files & directories containing the word.")
         (@arg file_input: -i --input +takes_value "Gives statistics from a previous tokei run. Can be given a file path, or \"stdin\" to read from stdin.")
         (@arg files: -f --files "Will print out statistics on individual files.")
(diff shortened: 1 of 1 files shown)


m1el:
This doesn't actually do what I intended, need more testing.

Aaronepower:
I think it might be `+multiple` ?

m1el:
`+multiple` has the same result as `...`.
The problems is that parser consumes multiple arguments to `--exclude` option.
E.g. `tokei --exclude *.rs ./src/` will parse as `{exclude: ["*.rs", "./src/"]}`, not `{exclude: ["*.rs"], inputs: ["./src/"]}`.
I don't think there's an easy way around this, but I'm reading clap source code for hints.

m1el:
The problem is that `--help` will print the following:

```
USAGE:
    tokei.exe [FLAGS] [OPTIONS] [--] [input]...
```

Which, IMO, is misleading because it's reasonable to expect `tokei (OPTIONS = --exclude *.rs) (input = ./src)`.

m1el:
`+multiple number_of_values(1)` seems to give expected results.

Aaronepower:
What if I want to do something like `tokei ./src --exclude ./bin ./html` would that exclude both bin and html?

m1el:
You can exclude multiple directories by specifying `--exclude` option multiple times:

```
tokei ./src --exclude ./bin --exclude ./html
```

Which makes perfect sense to me.

PR #115, opened 2017-04-08
Support "filenames" key, mk. 2

kazimuth:
Reopening #109 with comments addressed. (It took a while to get back to it, I've been swamped with work.)

We now support:
- Makefiles
- Rakefiles
- Dockerfiles
- CMakeLists.txt
- SCons


src/language/language_type.hbs.rs
```diff
@@ -53,22 +53,44 @@ impl LanguageType {
...
+                        {{~/each}}
+                            => return Some({{~@key}}),
+                    {{~/if}}
+                {{~/each}}
+
+                _ => ()
+            }
+        } 
+
+        let extension = fs::get_extension(&entry)
```
Aaronepower: I think that extension matching should be first. Just so it's a clear set of precedences in determining the language.
```
(extension) > (filename) > (shebang)
```

kazimuth: That will unfortunately prevent CMake from working. If you think it's the right thing I can implement it, though.

Aaronepower: Why is that?

kazimuth: The canonical CMake filename is `CMakeLists.txt`, for whatever reason. Maybe to make it easier for Windows users to edit?

Similar cases are `Cargo.toml` and `package.json`, although those are more straightforwardly Cargo and JSON files, so I understand why you wouldn't want to count them separately.

Note that the `filename` key matches the entire filename, not just the extension-stripped stem. So in this scheme, `Makefile.bananas` won't be detected as anything, since there's nothing with the `Makefile.bananas` filename, and nothing with the `bananas` extension.

Aaronepower: Why did you choose the entire filename over the stem?

kazimuth: In order to support CMakeLists.txt, mainly.

Aaronepower: Alright, seems fine with me, if you could rebase it onto the latest I'll be happy to merge it in.

kazimuth:
Rebased!

PR #135, opened 2017-07-30
Add tests for python2 and html files

si1kdd:
Hello, 

The comments features of tokei is still buggy.

I clone the newest version of tokei, than build by cargo, version 6.0.2

And I wrote 2 tiny scripts,
First one is for python, the result of tokei is

```
 Python                  1           18            8            5            5
 Total                   1           18            8            5            5
```
But actually should be 9 comment of lines.

and the html result is:
```
 HTML                    1           26           11           11            4
 Total                   1           26           11           11            4
```

But actually should be 7 comment of lines.
My enviroment is Bash and Linux x86_64.

Also, there are some flags buggy related to Zsh, and Bash.

I may create a issue later.

Thanks.


Aaronepower:
Thank you for your PR, and congrats on your first time contributing! I've now written a fix for the html file. The python test file is not correct as `"""Hello"""` is a string and not a comment. If you could remove the python file I'd be happy to merge it in.

si1kdd:
Ok, I delete the python file, 
but actually people would like to use doc string to do multiline comment, 
I think doc string should be included in comment ?  just my opinion :)
Sorry about that i am still a Rust rookie, couldn't help you to fix the features.

Aaronepower:
@si1kdd It's no worries. I understand that people use python doc strings as comments, however what it would require a lot more effort and special casing as I would need to know if it is being used as a doc string or just as a multi line string. If you'd like to learn Rust I'd recommend [reading the book](https://doc.rust-lang.org/stable/book/second-edition/).

PR #147, opened 2017-09-29
Use eprintln!

equal-l2:
`eprint!` and `eprintln!` has been introduced on Rust 1.19.0.
They should provide same functionalities as `err!` and `errln!`.


Aaronepower:
Thank you for your PR. Congratulations on your first contribution!

PR #169, opened 2018-01-03
add support for Meson

ignatenkobrain:
Fixes: https://github.com/Aaronepower/tokei/issues/168
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,6 +334,7 @@ Lucius
 Madlang
 Makefile
 Markdown
+Meson
 ModuleDef
 Mustache
 Nim
--- a/build.rs
+++ b/build.rs
@@ -28,7 +28,7 @@ fn generate_languages(out_dir: &OsStr) {
         serde_json::from_reader(json).expect("Can't parse json")
     };
 
-    for (key, ref mut item) in json.get_mut("languages")
+    for (_key, ref mut item) in json.get_mut("languages")
                                    .unwrap()
                                    .as_object_mut()
                                    .unwrap()
--- a/languages.json
+++ b/languages.json
@@ -816,6 +816,19 @@
             "extensions": ["def"],
             "line_comment": [";"]
         },
+        "Meson":{
+            "line_comment":[
+                "#"
+            ],
+            "quotes":[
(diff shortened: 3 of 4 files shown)


ignatenkobrain:
I tried to add test for it, but it fails...

```meson
# 12 lines 6 code 2 comments 4 blanks

project('xyz', 'c',
        meson_version : '>=0.30.0') # not counted

xyz_gen = '''
# comment inside
print("This is generated source.")
'''

# this is a comment

```

It treats `# comment inside` as a real comment while it is in multiline quote, so it seems to be bug in tokei.

Aaronepower:
@ignatenkobrain Can you add the test back in? So it can be added with the PR and I can test it.

ignatenkobrain:
@Aaronepower done.

Aaronepower:
@ignatenkobrain Can you rebase this onto the new master and see if the test continues to fail?

ignatenkobrain:
@Aaronepower 

```
---- meson stdout ----
	Meson LINES MATCH
thread 'meson' panicked at 'assertion failed: `(left == right)`
  left: `6`,
 right: `4`', /home/brain/Projects/upstream/tokei/target/debug/build/tokei-623c5e7e142d4c32/out/tests.rs:744:13
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Aaronepower:
@ignatenkobrain I've looked into the error and I think the problem is that you've set `'''` as a multi line comment when it is a string literal and should be in `quotes`, switching it to that makes it work.

ignatenkobrain:
@Aaronepower so I pushed new version, but it still fails because of lines of code, it thinks that it is 5, but I think that it is 6.

Aaronepower:
@ignatenkobrain Okay, the latest commit should fix that inaccuracy.

ignatenkobrain:
@Aaronepower thanks! should be ready for merge

Aaronepower:
@ignatenkobrain Thank you for the PR!

PR #170, opened 2018-01-06
Fix broken copyright link in ToC

seiks:
Self explanatory.


Aaronepower:
Thank you for this PR!

PR #171, opened 2018-01-06
Update crossbeam to 0.2.12

stjepang:
Crossbeam 0.2.10 uses the old auto trait syntax and 0.2.12 switched to the new one. Tokei is a part of Rust's test suite and, since it depends on crossbeam 0.2.10, it is currently blocking the following PR, which removes the old auto trait syntax in favor of the new one: https://github.com/rust-lang/rust/pull/46480#issuecomment-355735400

Diff between versions 0.2.10 and 0.2.12: https://github.com/crossbeam-rs/crossbeam/compare/0.2.10...v0.2.12

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,7 +71,7 @@ dependencies = [
 
 [[package]]
 name = "crossbeam"
-version = "0.2.10"
+version = "0.2.12"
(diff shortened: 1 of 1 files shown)


Aaronepower:
@stjepang Thank you for this PR, and congrats on your first contribution!

PR #174, opened 2018-01-10
Add FreeBSD install instructions

t6:
Hi,

I'm maintaining the FreeBSD [tokei port](https://www.freshports.org/devel/tokei). It'd be nice if it could be mentioned in the README.

Thanks!


Aaronepower:
@t6 Thank you for making tokei available on FreeBSD, and congrats on your first contribution!

PR #178, opened 2018-01-17
fix build with features enabled

ignatenkobrain:
Since 8817762d7b8f5cd10886147b828f95d19dc2a80a, build fails with any
features enabled..

```rust
error[E0308]: mismatched types
  --> src/main.rs:95:30
   |
95 |         match_output(format, &languages);
   |                              ^^^^^^^^^^
   |                              |
   |                              expected struct `tokei::Languages`, found reference
   |                              help: consider removing the borrow: `languages`
   |
   = note: expected type `tokei::Languages`
              found type `&tokei::Languages`
```

Problem is that when no features enabled, match_output() doesn't need
to "eat" `tokei::Languages`, but in real version of match_output it
must not be borrowed.

Fixes: https://github.com/Aaronepower/tokei/issues/177
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -171,7 +171,8 @@ mod io {
         process::exit(1);
     }
 
-    pub fn match_output(_format: &str, _languages: &Languages) -> ! {
+    #[cfg_attr(feature = "cargo-clippy", allow(needless_pass_by_value))]
+    pub fn match_output(_format: &str, _languages: Languages) -> ! {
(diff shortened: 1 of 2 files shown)


Aaronepower:
Thank you for this PR!

Aaronepower:
Thank you for this PR!

PR #181, opened 2018-01-25
supported vue files

fengcms:
supported vue files

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -1304,6 +1304,18 @@
                 "vim"
             ]
         },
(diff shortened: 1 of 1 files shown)


Aaronepower:
@fengcms Thank you for this PR! Could you add a test file to `tests/data` which should just be a vue file with a comment at the top saying how many lines of code, comments, blanks there are. If you need help making it there are plenty of other files there that you can sample from and if you're stuck feel free to ask me.

PR #191, opened 2018-02-15
Refactor how compiled-in serialization formats are handled

Ryman:
Went a bit deeper on this than planned, I'm trying to work towards having a `--stream-output` option so I can use tokei to just stream me a bunch of filenames with their stats (not interested in the summary) but I got stuck refactoring a bit of stuff as I was trying to understand how the library was working. 

The previous method of handling formats was quite hard to know when you've missed a method if you want to add a new type and it wasn't easy to get the error messages correct. In theory now someone can use tokei as a library and serialize with whatever format they want using the serde traits directly.

I'm probably going to continue onto making changes to the main library but I felt like this was a good point to check in to see if what I've done already would be wanted upstream or not :) 

I did notice there's some weird behavior currently that is broken even on a cargo-installed version that I'm not sure how you want it to behave: 
- what is `--input` meant to actually do, it never seems to add results to a scan when I try it?
- is there a limitation on toml support? I can't find any directory that it works on without giving me a 'key was not a string' error

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -127,16 +127,7 @@ supported_formats!(
             extern crate serde_cbor;
             extern crate hex;
 
-            use std::error::Error;
-            use std::process;
-
-            let hex: Vec<u8> = match hex::FromHex::from_hex(input) {
-                Ok(hex) => hex,
-                Err(err) => {
-                    eprintln!("{}", err.description());
-                    process::exit(1)
-                }
-            };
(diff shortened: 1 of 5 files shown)


Ryman:
Seems like windows is broken for the gnu builds... not sure if that's due to something I've done?

Aaronepower:
@Ryman Thank you for this PR, those breaks are not a result of this PR rather a problem with a dependency that we can't update yet.

Aaronepower:
Thanks again for this PR!

PR #192, opened 2018-02-20
Update to rayon 1.0

cuviper:


cuviper:
I didn't see that #190 already updated rayon, along with many other deps.  Feel free to close this if you want their larger update.

Aaronepower:
I am going to use that one as they got there first.  Thank you for this PR though!

PR #197, opened 2018-02-24
Fix typos. Add Scheme.

sprang:
Also "LISP" should be "Lisp".


Aaronepower:
Thank you for your PR and congrats your first contribution!

PR #204, opened 2018-03-10
Added Kakoune script.

notramo:
http://kakoune.org


tests/data/kakoune_script.kak
```diff
@@ -0,0 +1,13 @@
+# 13 lines, 8 code, 2 comments, 3 blank
```
Aaronepower: Should be `blanks` instead of `blank`

Aaronepower:
Thank you for this PR! I've left a comment on the diff that should fix the builds failing.

notramo:
@Aaronepower I have corrected it. Is it good now?

Aaronepower:
Thank you for this PR!

PR #212, opened 2018-04-07
Add Python identifiers for the env command

seiks:
Hello! :wave:

At the moment, Python scripts without an extension get ignored by Tokei, so if you have scripts taking the role of executables in the same fashion as shell scripts, Tokei won't count those files properly.

So, this PR will make no-extension files with this line of code work as expected for end-users:

```
#!/usr/bin/env python3
```

This PR adds identifiers for **python**, **python2** and **python3**.


seiks:
Done. Waiting for Travis.

seiks:
Any idea on why the NetBSD build failed?

Aaronepower:
Thank you for this PR!

PR #216, opened 2018-04-10
update README.md with how install output features

nacimgoura:
Informs the user that additional dependencies must be installed to use the output features. Without this, the user may think that all serialization formats are included without additional installation.


Aaronepower:
Thank you for your PR and congrats on your first contribution!

PR #219, opened 2018-04-16
Add Racket to the supported languages.

aatxe:
This PR would add Racket as a supported language for tokei. It's set up as a copy of the Scheme entry, but with the extension changed to `rkt`. The test file is changed to be valid Racket code.

A complication here is that line 17 of the test file _is_ a valid comment, but it comments out the whole s-expression regardless of how many lines it takes up. I don't believe this is currently possible with tokei, and so this PR does not currently count that line as a comment.

### DIFF
--- a/tests/data/racket.rkt
+++ b/tests/data/racket.rkt
@@ -1,4 +1,4 @@
-;;; 40 lines 14 code 15 comments 11 blanks
+;;; 40 lines 15 code 14 comments 11 blanks
 #lang racket ; defines the language we are using
 
 ;;; Comments
@@ -12,8 +12,8 @@
     |#
 |#
 
-;; S-expression comments discard the following expression,
-;; useful to comment expressions when debugging
+;; S-expression comments discard the following expression
+;; since this is syntax-aware, tokei counts this as code
 #; (this expression is discarded)
 
 ;; Constant
--- a/tests/data/racket.rkt
+++ b/tests/data/racket.rkt
@@ -1,0 +1,40 @@
+;;; 40 lines 14 code 15 comments 11 blanks
+#lang racket ; defines the language we are using
+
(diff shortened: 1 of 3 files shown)


tests/data/racket.rkt
```diff
@@ -0,0 +1,40 @@
...
+#| Block comments
+   can span multiple lines and...
+    #|
+        they can be nested!
+    |#
+|#
+
+;; S-expression comments discard the following expression,
+;; useful to comment expressions when debugging
+#; (this expression is discarded)
```
Aaronepower: This line is being counted as code. Is `#` or `#;` a comment syntax? If so it is missing from the definition and is causing your file to fail the tests.

Aaronepower:
Thank you for your PR! I've left a comment on the test, once you've addressed that issue and rebased onto the current master I'd be happy to merge it in.

tests/data/racket.rkt
aatxe: I mentioned this in the PR body, but `#;` comments out whole s-expressions, which I don't think is something tokei can handle right now. Maybe it makes sense to just count this as code anyway because it's used to comment out blocks of code quickly, rather than to write informative comments. Should I just update the test to not count this as a comment?

Aaronepower: @aatxe It's up to you. You could also remove this line since I don't see how tokei could ever really count this type of comment in a language agnostic way.

aatxe: Indeed, it definitely couldn't (since this is AST aware). I'll leave it, but update the test (and add a comment that this is counted as code by tokei).

Aaronepower:
@aatxe That is unrelated, thank you for your PR and congrats on your first contribution!

PR #228, opened 2018-05-21
Add Forsyth-Edwards-Notation (FEN) support

DenialAdams:
> A FEN "record" defines a particular game position, all in one text line and using only the ASCII character set. A text file with only FEN data records should have the file extension ".fen"

[FEN on Wikipedia](https://en.wikipedia.org/wiki/Forsyth%E2%80%93Edwards_Notation)

Fairly niche, and not code, so I understand perfectly if you opt to close this PR.

You can see an example file [here](https://github.com/DenialAdams/chessatk/blob/master/tests/positions.fen)

Thanks for your work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,6 +334,7 @@ Elm
 Emacs Development Environment
 Emacs Lisp
 Erlang
+FEN
 Forth
 F*
 F#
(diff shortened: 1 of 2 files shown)


Aaronepower:
Thank you for this PR! I feel like text based data formats are fine to add since we already include JSON. Congrats on your first contribution!

PR #235, opened 2018-06-12
Added support for XSLT

BrandonBoone:
`.xslt` is another common extension for `.xsl` documents. Not sure if you want a separate test for this or not as it will be identical to `.xsl`, but I added one just in case.

> XSLT documents are most often identified with the extensions ".xsl" or ".xslt".
[src](https://www.w3.org/TR/xslt20/)


Aaronepower:
I don't think adding an extension requires a test file since if one works the others will work too from how the code is generated. If you could remove that extra test I'd be happy to merge it in!

Aaronepower:
Thank you for your PR!

Issue #160, opened 2017-11-29
--files truncates paths at 24 characters


chris-morgan:
Tokei seems to try to fit its output in 79 columns, regardless of the terminal configuration (my terminal is 120 columns wide) or whether it’s being piped to something (at which point I would generally expect it to throw away any capping of line width, but wouldn’t quibble if it didn’t provided there was a command line argument).

Platform: Windows, Command Prompt.

The primary effect of this is that `tokei --files` truncates paths, showing only the last 24 characters of them; take this example in Rust’s libstd, for example:

```
C:\code\rust\src\libstd>tokei -f
-------------------------------------------------------------------------------
 Language            Files        Lines         Code     Comments       Blanks
-------------------------------------------------------------------------------
 Rust                  254        86150        48404        29682         8064
-------------------------------------------------------------------------------
[…]
 .\sys\windows\ext\ffi.rs           147           29          110            8
 .\sys\windows\ext\fs.rs            488           84          379           25
 .\sys\windows\ext\io.rs            209          136           52           21
 .\sys\windows\ext\mod.rs            44           21           19            4
 |s\windows\ext\process.rs          119           84           19           16
 .\sys\windows\ext\raw.rs            21            7           10            4
 |ys\windows\ext\thread.rs           31           16           10            5
 |\redox\net\dns\answer.rs           22           11            9            2
 |sys\redox\net\dns\mod.rs          215          177            9           29
 |s\redox\net\dns\query.rs           18            7            9            2
 |edox\syscall\arch\arm.rs           83           58           10           15
 |edox\syscall\arch\x86.rs           83           58           10           15
 |x\syscall\arch\x86_64.rs           84           59           10           15
 |trace\printing\dladdr.rs           45           32            9            4
 |acktrace\printing\mod.rs           43           28            9            6
 |\tracing\backtrace_fn.rs           48           25           19            4
[…]
```

This renders the output useless for machine reading.

I see three parts to fixing this:

1. Make it use the actual terminal width rather than a blind 79 columns (if it’s already supposed to be doing that, then change this to “detect console width on Windows”);
2. Add a command-line argument to uncap the width (like `ps -ww`);
3. Decide what to do when it is piped or the width cannot be determined (for reference, `ps` says of such cases that “the output width is undefined (it may be 80, unlimited, determined by the TERM variable, and so on)”, which is not necessarily a good situation).

Aaronepower:
[...]

PR #246, opened 2018-08-16
Acknowledge terminal width for output on Unix-systems

Veykril:
This PR implements #160 for the unix family, changing the output width from 79 to the terminal width if it is wider than 79. I pulled in libc as a dependency for the syscall to make. I will implement the windows variant as well, but I wanted to know if it is okay to use the winapi crate for this or if we should define the one or two functions that are required in the `main.rs` ourself.

Also this PR changes the default behaviour. I wasnt sure if this should be an argument or not. If it should be I'd gladly add that as well if you propose a definition for it.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,6 +20,7 @@ use tokei::Sort;
 use input::Format;
 
 const FILES: &str = "files";
+const FALLBACK_ROW_LEN: usize = 79;
 const NO_LANG_ROW_LEN: usize = 61;
 const NO_LANG_ROW_LEN_NO_SPACES: usize = 54;
 
@@ -178,7 +179,7 @@ fn main() {
         process::exit(0);
     }
 
-    let columns = terminal_columns().max(79);
+    let columns = terminal_columns().max(FALLBACK_ROW_LEN);
     let row = "-".repeat(columns);
 
     println!("{}", row);
@@ -272,7 +273,7 @@ fn terminal_columns() -> usize {
         dwMaximumWindowSize: COORD { X: 0, Y: 0 },
     };
     if unsafe { GetConsoleScreenBufferInfo(GetStdHandle(STD_OUT_HANDLE), &mut buffer_info) } == 0 {
-        79
+        FALLBACK_ROW_LEN
     } else {
         (buffer_info.srWindow.Right - buffer_info.srWindow.Left + 1) as usize
     }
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -201,35 +203,37 @@ fn main() {
...
+        println!("{}", row);
     }
     let mut total = Language::new_blank();
     for (_, language) in languages {
         total += language;
     }
 
-    print_language(&total, "Total");
-    println!("{}", ROW);
+    print_language(columns - 61, &total, "Total");
```
Aaronepower: What is 61? This should probably be a `const` instead of a magic number.

src/main.rs
```diff
@@ -201,35 +203,37 @@ fn main() {
...
+    println!("{}", row);
 }
 
-fn print_results<'a, I>(languages: I, list_files: bool)
+fn print_results<'a, I>(row: &str, languages: I, list_files: bool)
 where I: std::iter::Iterator<Item = (&'a LanguageType, &'a Language)> {
+    let path_len = row.len() - 54;
```
Aaronepower: Same as above about magic numbers.

Veykril:
Added a winapi version and changed the magic numbers to constants. I hope the names for them are okay cause I wasnt really sure what to call them. They both are the number of characters in a row that exclude the language/path section with and without the spaces between the sections counted. Maybe you can come up with a better name for them, and on that note they probably shouldnt contain lang but path in their names instead.

src/main.rs
```diff
@@ -176,16 +179,19 @@ fn main() {
...
                 "Language",
                 "Files",
                 "Lines",
                 "Code",
                 "Comments",
-                "Blanks");
-    println!("{}", ROW);
-
+                "Blanks",
+                columns - 67);
```
Aaronepower: There is another magic number here.

Aaronepower:
Hmm, looking at this code with the `winapi` version makes me wonder if `terminal_columns` should actually be a seperate crate, as I think this would help a lot of people writing CLI apps. How would you feel about pulling out that code and making it a crate?

Veykril:
Sounds like a good idea to explore, a small crate with helper functions like that. I'll experiment with that a bit and come back to this later.

Veykril:
After searching a bit more it turns out there is already such a [crate](https://github.com/clap-rs/term_size-rs), I changed the implementation to rely on this instead.

src/stats.rs
```diff
@@ -72,13 +73,15 @@ impl fmt::Display for Stats {
...
-        if name_length == 25 || name_length <= 24 {
-            display_stats!(f, self, name)
+        let max_len = f.width().unwrap_or(25);
+
+        if name_length <= max_len {
+            display_stats!(f, self, name, max_len)
         } else {
             let mut formatted = String::from("|");
-            let from = find_char_boundary(&name, name_length - 24);
+            let from = find_char_boundary(&name, name_length + 1 - max_len);
```
Aaronepower: Can you put comment here explaining the `+ 1`?

Veykril:
Does this explain it properly?

Aaronepower:
@Veykril Yeah looks good to me, once the CI has completed I'll merge it in.

PR #258, opened 2018-09-05
Add percent column

BrainMaestro:
Fixes #230

### DIFF
--- a/src/language/mod.rs
+++ b/src/language/mod.rs
@@ -106,3 +106,12 @@ impl AddAssign for Language {
     }
 }
 
+impl<'a> AddAssign<&'a Self> for Language {
+    fn add_assign(&mut self, rhs: &Self) {
+        self.lines += rhs.lines;
+        self.comments += rhs.comments;
+        self.blanks += rhs.blanks;
+        self.code += rhs.code;
+        self.stats.extend(rhs.stats.clone());
+    }
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -110,6 +110,8 @@ fn main() -> Result<(), Box<Error>> {
             or \"stdin\" to read from stdin.")
         (@arg files: -f --files
             "Will print out statistics on individual files.")
+        (@arg percent: -p --percent
+            "Will print out regular statistics with percentages related to the rest of the project.")
         (@arg input:
             conflicts_with[languages] ...
             "The input file(s)/directory(ies) to be counted.")
@@ -149,6 +151,7 @@ fn main() -> Result<(), Box<Error>> {
         }
         ignored_directories
     };
+    let percent_option = matches.is_present("percent");
 
     // Sorting category should be restricted by clap but parse before we do work just in case.
     let sort_category = sort_option.map(parse_or_exit::<Sort>);
@@ -185,6 +188,11 @@ fn main() -> Result<(), Box<Error>> {
     });
 
     languages.get_statistics(&paths, ignored_directories, types);
+    let mut total = Language::new();
+
+    for (_, language) in &languages {
+        total += language;
+    }
 
     if let Some(format) = output_format {
         print!("{}", format.print(languages).unwrap());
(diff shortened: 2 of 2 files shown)


src/language/mod.rs
```diff
@@ -106,3 +117,20 @@ impl AddAssign for Language {
     }
 }
 
+impl<'a> Add<&'a Self> for Language {
+    type Output = Self;
+
+    fn add(self, rhs: &Self) -> Self {
+        let mut stats = self.stats;
+        stats.extend(rhs.stats.clone());
```
BrainMaestro: The `clone` here is unfortunate, the only way I see around it is for stats to be `Vec<Rc<Stat>>`, do you have a better solution?

BrainMaestro:
I decided not to add a test for this because it was basically pointless since percentage is 100% for the individual files.

src/language/mod.rs
```diff
@@ -29,6 +30,8 @@ pub struct Language {
     pub lines: usize,
     /// A collection of statistics based on the files provide from `files`
     pub stats: Vec<Stats>,
+    /// Total percentage of the language in the project
+    pub percent: usize,
```
BrainMaestro: would you prefer a float?

Aaronepower:
Thank you for your PR! I'm sorry I should have given more guidance and said what I think the feature should be. As I didn't actually want to add a column. The way I saw this working was behind a flag. I also saw instead of and extra column I saw the output looking like below. Where it showed the absolute and relative values of each compared to the total. For the `—files` output they would be compared to the language.
```
--------------------------------------------------------
 Language                                        Files
--------------------------------------------------------
 Batch                                         1|0.29%
 C                                             1|0.29%
 Makefile                                      1|0.29%
 Markdown                                     19|5.5%
 Python                                       75|22%
 ReStructuredText                              7|2%
 Rust                                        210|61%
 Shell                                         7|2%
 TOML                                         17|5%
 Vim Script                                    2|0.58%
-------------------------------------------------------
 Total                                          340
-------------------------------------------------------
```

BrainMaestro:
That is indeed much better. I wasn't comfortable with the way it was implemented anyway. I'll fix

Aaronepower:
@BrainMaestro Thank you for working on this, and if you have any questions don't hesitate to ask.

BrainMaestro:
See what it looks like 
```---------------------------------------------------------------------------------
 Language              Files        Lines         Code     Comments       Blanks
---------------------------------------------------------------------------------
 JSON               1|23.26%         1622         1622            0            0
 Markdown           5|26.40%         1841         1841            0            0
 Rust              13|29.43%         2052         1557          207          288
 Shell              4|1.84%           128           92            8           28
 SVG                8|16.72%         1166         1126            0           40
 TOML               1|1.20%            84           69            0           15
 YAML               1|1.15%            80           60            8           12
---------------------------------------------------------------------------------
 Total                    33         6973         6367          223          383
---------------------------------------------------------------------------------
```

Should the total files be aligned with the files not the percentages?

Aaronepower:
@BrainMaestro I think aligned with the number but I'll have to see to confirm.

src/main.rs
```diff
@@ -277,17 +279,61 @@ fn isnt_empty(&(_, language): &(&LanguageType, &Language)) -> bool {
...
+                 language.lines,
+                 language.code,
+                 language.comments,
+                 language.blanks,
+                 len = lang_section_len)
+    }
+
+    let percent_char_count = 5;
+    // don't show percentage for the total row because that would be pointless
+    if language.lines == total.lines {
```
Aaronepower: Instead of passing in total and performing this if statement, why not just call ` print_language(&mut stdout, columns - NO_LANG_ROW_LEN, &total, &total, "Total", percent_option)?;` as ` print_language(&mut stdout, columns - NO_LANG_ROW_LEN, &total, &total, "Total", false)?;`?

Aaronepower: I think then that there should probably be a seperate function for printing total. Currently this implementation changes a lot of behaviour (such as moving the total around which adds additional allocations in `AddAssign`) which is not desired.

BrainMaestro: How would you suggest I do that?

Aaronepower: So I've refactored all the printing code into `src/cli_utils.rs` so you'll need to move your changes to that. I would suggest having a function called `print_total` that handles all the logic of printing out the total.

BrainMaestro: I'm not able to rework this now, so I'll pick it up again later when I can. I'll just go ahead and close the PR until then. thanks

Aaronepower: @BrainMaestro Okay, thank you again for your PR! I'm going to take sometime and write a comment on the issue detailing exact requirements so that if and when you do get back to it you'll have a much clearer sense of the task.

BrainMaestro: @Aaronepower that sounds great.

fnzac: > I'm going to take some time and write a comment on the issue detailing exact requirements

Hi, was this ever written? I'm interested in trying to (re)implement this. Although I'm new to Rust and open source.

omid:
@XAMPPRocky I couldn't find the original issue!
But I think a separated column is better, specially for other formats like json.
Anyway it's a piece of data that can be useful for other services.

PR #260, opened 2018-09-06
Liquid template language support

t-richards:
Adds support for the Liquid template language. See also:

 - https://shopify.github.io/liquid/
 - https://github.com/Shopify/liquid

Test data file copied from the Liquid project itself, licensed MIT.

https://github.com/Shopify/liquid/blob/master/performance/tests/dropify/collection.liquid

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -881,6 +881,20 @@
                 ["'", "'"]
             ]
         },
+        "Liquid": {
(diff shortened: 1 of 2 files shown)


Aaronepower:
@t-richards Thank you for your PR and congrats on your first contribution!

PR #262, opened 2018-09-10
Add man page

The-King-of-Toasters:
A simple translation of the help command. I removed/changed some of the text to make it more `man`-like.


Aaronepower:
@The-King-of-Toasters Thank you for this work, however I'm not currently willing to maintain the manpage as versions are released and `clap` has plans to have man pages generated automatically so I would prefer to hold off until such a feature is availble.

Issue #165, opened 2017-12-29
Ugly and unusable error message when permission denied for a file/directory


notramo:
The error message of `tokei` is ugly, when it isn't permitted to read a directory/file.
directory:
```
ERROR:tokei::utils::fs: permission denied
```

file:
```
ERROR:tokei::language::languages: permission denied
```

The error message looks like it is meant for the developer, and it is not too much usable for the user.
It *doesn't contain a filename*, instead it contains unnecessary information about how the file/directory reading is internally implemented.

Expected behaviour:

- directory: `Permission denied for reading /path/to/dir`
- file: similar to the directory error message, but if the extension is known, mark the language in the final result:
  ```
  Permission denied for reading /path/to/file.rs
  Permission denied for reading file.cr
  Note: results can be inaccurate for languages marked with '(!)'
  -------------------------------------------------------------------------------
   Language            Files        Lines         Code     Comments       Blanks
  -------------------------------------------------------------------------------
   Crystal (!)            14         1144          945           54          145
   YAML                    2          121           77           25           19
   Nim                     2          193          136           23           34
   Rust (!)               16         2102         1405          456          241
   D                       1            8            5            1            2

PR #270, opened 2018-10-01
Show Inaccuracy Warning on permission error

Veykril:
This PR roughly implements issue #165. I am sure there is room for some improvements on my code changes so feel free to comment on everything I changed. One thing I dislike that I did is this weird if let bundle I created to check for a permission error at one place:
https://github.com/Veykril/tokei/blob/0c7e18d1b0b45832c40c3a991be4bd57111ff5e5/src/utils/fs.rs#L52-L61
I couldnt come up with a better way since the nested errors are boxed and box patterns arent stable yet.

Example Output:
```
C:\Workspace\Rust\tokei (error_output -> origin)                                                                       
λ cargo run -- C:\Workspace\Rust\test                                                                                  
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s                                                          
     Running `target\debug\tokei.exe C:\Workspace\Rust\test`                                                           
ERROR 2018-10-01T16:28:56Z: tokei::utils::fs: Permission denied for reading C:\Workspace\Rust\test\Cargo.toml          
Note: results can be inaccurate for languages marked with '(!)'                                                        
-----------------------------------------------------------------------------------------------------------------------
 Language                                                    Files        Lines         Code     Comments       Blanks 
-----------------------------------------------------------------------------------------------------------------------
 Markdown                                                        4         1380         1380            0            0 
 Rust                                                            1          123           99            1           23 
 Shell                                                           1           40           29            1           10 
 TOML (!)                                                        1           84           69            0           15 
 YAML                                                            1           80           60            8           12 
-----------------------------------------------------------------------------------------------------------------------
 Total (!)                                                        8         1707         1637           10           60
-----------------------------------------------------------------------------------------------------------------------
```


Veykril:
@Aaronepower sorry for pinging you but I wasnt sure whether you didn't see this PR or not

Aaronepower:
@Veykril Sorry, yes I did see it. I have a problem where if I'm unsure about a feature I tend to not reply till I have a more definitive answer 😅. Could you run `benchmark.sh` in the root of the repo and post the output here?

Veykril:
I was concerned about the performance impact of this as well but i couldn't think of a different way to implement this.

```
Benchmark 1: tokei "C:WorkspaceRusttokei"  
Summary
Time (mean ± σ):      13.1 ms ±   2.1 ms    [User: 5.3 ms, System: 4.3 ms]
Range (min … max):    10.1 ms …  22.9 ms

Benchmark 2: tokei "C:WorkspaceRusttokei"
Time (mean ± σ):      14.2 ms ±   2.9 ms    [User: 4.2 ms, System: 5.5 ms]
Range (min … max):    10.0 ms …  26.0 ms
Summary
  'tokei.exe "C:WorkspaceRusttokei"' ran
    1.08 ± 0.28 times faster than 'tokei "C:WorkspaceRusttokei"'
```
I don't think these are accurate though, my machine isnt the best to benchmark things + Im running windows so im not too sure how well that goes.

Aaronepower:
@Veykril Sorry, could you run the benchmark on a larger repo such as the rust repo?

Veykril:
Okay I ran it on the rust repo and I moved and renamed my exe out of the build dir cause i wasnt sure how windows handles path exes and the exes in the current dir, so iwanted to be safe that it actually executes my build.

Output without an error: https://pastebin.com/3xHqpUJT
Output with an error: https://pastebin.com/ALFnSJqx

Aaronepower:
@Veykril Sorry about this, I've since made a large scale refactor to the codebase that makes the changes here incompatible. You'll need to rewrite some of this for it to be accepted.

Veykril:
No problem, will look into your changes sometime later this week and try to adapt my PR to that

Veykril:
Alright, I adapted the changes and tested them. From what I have seen I didnt have to change much unless I've overlooked something.

src/language/mod.rs
```diff
@@ -104,6 +112,9 @@ impl AddAssign for Language {
         self.blanks += rhs.blanks;
         self.code += rhs.code;
         self.stats.extend(mem::replace(&mut rhs.stats, Vec::new()));
+        if rhs.inaccurate {
+            self.inaccurate = rhs.inaccurate
```
Aaronepower:
```suggestion
            self.inaccurate |= rhs.inaccurate
```

src/language/language_type.rs
```diff
@@ -18,22 +18,26 @@ include!(concat!(env!("OUT_DIR"), "/language_type.rs"));
 impl LanguageType {
     /// Parses a given `Path` using the `LanguageType`. Returning `Stats`
     /// on success.
-    pub fn parse(self, path: PathBuf) -> io::Result<Stats> {
+    pub fn parse(self, path: PathBuf) -> Result<Stats, (io::Error, PathBuf)> {
```
Aaronepower: Instead of changing the API signature can we instead have a `parse_with_accuracy` or something similar that has this signature, and have `parse` call that and replace the `parse` call in `src/utils/fs.rs` with `parse_with_accuracy`?

src/language/language_type.rs
```diff
@@ -18,22 +18,26 @@ include!(concat!(env!("OUT_DIR"), "/language_type.rs"));
...
         };
 
-        self.parse_from_str(path, &text)
+        Ok(self.parse_from_str(path, &text))
     }
 
     /// Parses the text provided. Returning `Stats` on success.
-    pub fn parse_from_str(self, path: PathBuf, text: &str)
-        -> io::Result<Stats>
+    pub fn parse_from_str(self, path: PathBuf, text: &str) -> Stats
```
Aaronepower: Ditto on changing the API.

src/language/language_type.rs
```diff
@@ -19,23 +19,38 @@ impl LanguageType {
...
     /// Parses the text provided. Returning `Stats` on success.
-    pub fn parse_from_str(self, path: PathBuf, text: &str)
-        -> io::Result<Stats>
+    pub fn parse_from_str(self, path: PathBuf, text: &str) -> io::Result<Stats>
     {
+        Ok(self.parse_from_str_with_accuracy(path, text))
+    }
 
+    /// Parses the text provided.
+    pub fn parse_from_str_with_accuracy(self, path: PathBuf, text: &str) -> Stats
```
Aaronepower: Why is `parse_from_str_with_accuracy` necessary?

Veykril: oh it actually isnt now that you say it, i was just confused as to why `parse_from_str`returns a result when it never returns an error. Probably due to previous implementations I guess? Will remove the function again

src/utils/fs.rs
```diff
@@ -44,10 +45,20 @@ pub fn get_all_files(paths: &[&str],
...
         let tx = tx.clone();
         Box::new(move |entry| {
-
             let entry = match entry {
                 Ok(entry) => entry,
                 Err(error) => {
+                    use ignore::Error;
+                    if let Error::WithDepth { err: ref error, .. } = error {
+                        if let Error::WithPath { ref path, err: ref error } = **error {
+                            if let Error::Io(ref error) = **error {
```
Aaronepower: I think we can condense this code. Instead of just checking for Permission denied, couldn't we have `error!("{} reading {}", error.description(), path.display());`?

Veykril: Yep, this is indeed better, I guess I was too focused on the PermissionDenied error there. Looks a lot better like this and makes more sense overall

src/language/language_type.rs
```diff
@@ -19,23 +19,31 @@ impl LanguageType {
...
-            let f = File::open(&path)?;
+            let f = match File::open(&path) {
+                Ok(f) => f,
+                Err(e) => return Err((e, path)),
+            };
             let mut s = String::new();
             let mut reader = DecodeReaderBytes::new(f);
 
-            reader.read_to_string(&mut s)?;
+            if let Err(e) = reader.read_to_string(&mut s) {
```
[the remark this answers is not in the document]
Veykril: I wanted to do that but the problem is that the closure eagerly moves the pathbuf even when it doesnt get executed

src/language/language_type.rs
```diff
@@ -19,23 +19,31 @@ impl LanguageType {
...
+                return Err((e, path));
+            }
             s
         };
-
-        self.parse_from_str(path, &text)
+        Ok(self.parse_from_str(path, &text).unwrap())
```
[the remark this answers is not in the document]
Veykril: Oh right, I did this cause the results have different error cases and parse_from_str cannot fail from its implementation. Probably a bad idea to unwrap though since the implementation could change. Will have to do the same if let Err instead of map_err though due to the closure move that would happen otherwise though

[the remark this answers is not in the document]
Veykril: Actually i have a problem here in general cause the parse_from_str error that wouldnt happen in its current implementation doesnt give back the pathbuf, so I am not sure how to even map the error here

src/language/language_type.rs
```diff
@@ -19,23 +19,31 @@ impl LanguageType {
...
     pub fn parse(self, path: PathBuf) -> io::Result<Stats> {
+        self.parse_with_accuracy(path).map_err(|e| e.0)
+    }
+
+    /// Parses a given `Path` using the `LanguageType`. Returning `Stats`
+    /// on success and giving back ownership of PathBuf on error.
+    pub fn parse_with_accuracy(self, path: PathBuf) -> Result<Stats, (io::Error, PathBuf)> {
         let text = {
-            let f = File::open(&path)?;
+            let f = match File::open(&path) {
```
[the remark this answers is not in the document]
Veykril: same problem here with the closure movement

Veykril:
No problem, I am still learning you pointing all these things out is only helping me in the end. Better to do this properly than just letting everything through 👍

Aaronepower:
@Veykril Okay,  I think if you could make the style change and resolve the conflicts this is ready to merge in.

Aaronepower:
Haha, sorry you actually can break the API as I'm going to be making some changes that would also break the API so we might as well break them both in the same release.

Veykril:
So i can change the function signatures as I please?

Aaronepower:
@Veykril Remove the `accuracy` functions and change to the API of the current ones.

Veykril:
I suppose this is what you meant right? I hope

Aaronepower:
Okay, this PR looks good to me. Thank you again for your PR and for making the requested changes!

PR #271, opened 2018-10-03
Handling verbose argument

### LINKED ISSUE #242: -v 2 option is broken
Coming from: https://bugzilla.redhat.com/show_bug.cgi?id=1563136

```
[lnie@localhost kernel-tests]$ tokei -v 2 -o json
ERROR 2018-04-03T08:53:49Z: tokei::utils::fs: entity not found
{}[lnie@localhost kernel-tests]$ tokei -v 2 -o yaml
ERROR 2018-04-03T08:53:57Z: tokei::utils::fs: entity not found
---
{}[lnie@localhost kernel-tests]$ tokei -v 2
ERROR 2018-04-03T08:54:19Z: tokei::utils::fs: entity not found
 Total                   0            0            0            0            0

```

patricksjackson:
This changes the functionality of the `--verbose` flag to what seems to be intended (see #242). Before this change you need to specify the logging levels by the number of `-v`'s you added to the command.

For example, the old way of specifying logging level 3 would be `tokei -v -v -v`. This change makes the invocation `tokei -v 3`.

Closes #242

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -33,16 +33,16 @@ fn main() -> Result<(), Box<Error>> {
             "Gives statistics from a previous tokei run. Can be given a file path, \
             or \"stdin\" to read from stdin.")
         (@arg files: -f --files
-            "Will print out statistics on individual files.")
+            "Prints out statistics on individual files")
         (@arg input:
             conflicts_with[languages] ...
             "The input file(s)/directory(ies) to be counted.")
         (@arg types: -t --type
             +takes_value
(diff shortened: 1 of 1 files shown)


Aaronepower:
Thank you for your PR! However it is currently working as intended and it is the documentation that is wrong. If the change to code was removed and the documentation was made clearer I would accept that. It should probably be like the following.

```
-v: show unknown file extensions
-vv: reserved for future debugging
-vvv: enable file level trace, not recommended on
```

PR #272, opened 2018-10-03
Let --sort argument be case insensitive.

patricksjackson:
Allows the `--sort` argument to be case insensitive. Since the column names are capitalized in the output, this makes it a little more natural so that you can actually use the printed column names as the argument to sort.


Aaronepower:
Thank you for PR, and congrats on your first contribution!

Issue #62, opened 2016-09-11
Handle variable comments


Aaronepower:
Stuff like the following should 3 lines of comments.

```
<<<END
comment
END
```

Aaronepower:
Actually looking into this, these are treated universally as strings, so they are not needed to be counted as comments, but rather should go into how we handle looking for comments in strings.

dolmen:
The label "bug" for this issue would be more appropriate than "enhancement".
Also, the README claims that "Tokei is accurate". So, is this issue fixed?

Aaronepower:
@dolmen It isn't a bug, as that would imply that it did count heredoc strings before. Tokei is accurate, token is *not* perfectly accurate. Without writing parsers for every language tokei will always have edge cases it can't cover.

PR #274, opened 2018-10-11
Add language rules for Perl6 and its cousin Not Quite Perl

alekratz:
Perl6 is just slightly different from Perl 5, with the multline comments starting/ending with `=begin` and `=end` instead of `=pod` and `=cut`.

I have not written Not Quite Perl at all, but it is the language that Perl6 is written in and has very similar syntax, and effectively identical comment syntax.

The tests for these are not entirely accurate. Perl6 allows for multiline comments to have a "tag" of sorts, e.g.

```perl6
=begin item match
Documentation for match here
=end item
```

where "item" is the tag. While one would normally include the `=end item` to be part of the multiline comment, tokei only ends at the `=end` token. Is it possible to define regular expressions for multiline comments? I have not looked too closely at the source.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -1092,6 +1092,22 @@
                 "pm"
             ]
         },
+        "Perl6":{
+            "line_comment":[
+                "#"
+            ],
+            "multi_line":[
+                ["=begin", "=end"]
+            ],
+            "quotes":[
+                ["\\\"", "\\\""],
+                ["'", "'"]
+            ],
+            "extensions":[
+                "pl6",
+                "pm6"
+            ]
+        },
         "Php":{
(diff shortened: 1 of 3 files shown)


Aaronepower:
Thank you for your PR! Yes, I've also written some Perl6 and the reason it wasn't included was because of how hard it is to correctly count its syntax. Not just with heredoc strings / comments (#62) but with quoting constructs and how you can create arbitrary string literals.

Tokei does not use regular expressions to count code. Though bringing it up in this context I could see an implementation for capturing heredoc style literals.

I think this is fine for now until there is a solid solution for heredocs. Congrats on your first contribution!

PR #276, opened 2018-10-16
add support for wavefront obj and mtl formats

expenses:
Hi, feel free to disregard this if its just bloat, but I added language configurations for [Wavefront OBJ](https://en.wikipedia.org/wiki/Wavefront_.obj_file) files and [Wavefront MTL](https://en.wikipedia.org/wiki/Wavefront_.obj_file#Material_template_library) files.


Aaronepower:
@expenses Thank you for your PR! These look fine to me, can you add a test file for each in `tests/data` and rebase this onto the latest commit? I just fixed the CI issue.

expenses:
Ok, I believe I did that correctly. The obj/mtl files I used were just the default exported cube from blender.

Aaronepower:
@expenses The first line should contain of human verified number of lines, comments, blanks, and code. Check out the [Rust test case](https://github.com/Aaronepower/tokei/blob/master/tests/data/rust.rs) for an example, and if you need any help with this feel free to ask!

PR #279, opened 2018-11-09
Access files to analyze through FileAccess trait

udoprog:
This pushes the abstractions available to provide files for analysis down a bit, and exposes a refactored, lower-level version of `get_statistics` that allows you to provide an iterator of `FileAccess` objects for analysis.

This allows you to analyze non-filesystem like things, like tar and zip archives. Or in-memory collections of files by building an implementation of `FileAccess`.

A simple example would look something like (warning: not tested):

```rust
use std::io;

struct FakeFile {
    name: String,
    content: String,
}

impl<'a> FileAccess<'a> for &'a FakeFile {
    fn read_to_string(self) -> io::Result<String> {
        Ok(self.content.clone())
    }

    fn read_first_line(self) -> io::Result<String> {
        let line = self.content.split("\n").next().map(|s| s.to_string()).unwrap_or_else(|| self.content.to_string());
        Ok(line)
    }

    fn name(self) -> Cow<'a, str> {
        Cow::from(&self.name)
    }

    fn file_name(self) -> Option<Cow<'a, str>> {
        Some(self.name.rsplit("/").next().map(Cow::from).unwrap_or_else(|| Cow::from(&self.name)))
    }

    fn extension(self) -> Option<Cow<'a, str>> {
        match self.file_name() {
            Some(Cow::Borrowed(n)) => n.rsplit(".").next().map(Cow::from),
            Some(Cow::Owned(n)) => n.rsplit(".").next().map(|s| s.to_string()).map(Cow::from),
            None => None,
        }
    }
}
```

### DIFF
--- a/src/file_access.rs
+++ b/src/file_access.rs
@@ -53,7 +53,7 @@ impl<'a> FileAccess<'a> for &'a Path {
 /// It can be converted into a LanguageType (e.g. identify which language it belongs to).
 /// The name of the file is typically its path, but this might be logical in case it's part of an
 /// archive (tar, zip, ...).
-pub trait FileAccess<'a>: Copy + Send {
+pub trait FileAccess<'a>: Copy {
     /// Read the contents of the file into a string.
     fn read_to_string(self) -> io::Result<String>;
 
--- a/src/file_access.rs
+++ b/src/file_access.rs
@@ -1,0 +1,71 @@
+use std::io;
+use encoding_rs_io::DecodeReaderBytes;
+use std::fs;
+use std::path::Path;
+use std::borrow::Cow;
+
+impl<'a> FileAccess<'a> for &'a Path {
+    fn read_to_string(self) -> io::Result<String> {
+        use std::io::Read;
+
+        let f = fs::File::open(self)?;
+        let mut s = String::new();
+        DecodeReaderBytes::new(f).read_to_string(&mut s)?;
+        Ok(s)
+    }
+
+    fn read_first_line(self) -> io::Result<String> {
+        use std::io::{BufReader, BufRead};
+
+        let file = fs::File::open(self)?;
+        let mut buf = BufReader::new(file);
+        let mut line = String::new();
+        let _ = buf.read_line(&mut line);
+
+        Ok(line)
+    }
+
+    fn name(self) -> Cow<'a, str> {
+        self.to_string_lossy()
+    }
+
+    fn file_name(self) -> Option<Cow<'a, str>> {
+        match self.file_name() {
+            Some(filename_os) => {
+                Some(Cow::from(filename_os.to_string_lossy().to_lowercase()))
+            },
+            None => None
+        }
+    }
+
+    fn extension(self) -> Option<Cow<'a, str>> {
+        match self.extension() {
+            Some(extension_os) => {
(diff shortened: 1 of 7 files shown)


Aaronepower:
Thank you for this PR! Is there a specific use case you're using this for? While this abstraction is interesting currently this doesn't provide any immediate benefit. You mention accessing tar and zip files, I would like to provide that but this doesn't currently provide that for the binary.

While looking through this I was wondering what about this requires being in `tokei` itself, and not in its own crate. If there was a crate that allowed access to zips and tars in the same way that [ignore's `Walk`](https://docs.rs/ignore/0.4.4/ignore/struct.Walk.html) struct does.

Lastly for the use case of in memory collections, I believe you can use tokei currently to read in memory collections using `parse_from_str`, as a PathBuf can be entirely faked and it won't attempt to access the filesystem.

udoprog:
> Thank you for this PR! Is there a specific use case you're using this for? While this abstraction is interesting currently this doesn't provide any immediate benefit. You mention accessing tar and zip files, I would like to provide that but this doesn't currently provide that for the binary.

I'm building that support in [another project](https://github.com/udoprog/kernelstats) to analyze usage statistics without having to unpack `tar.gz` files. That's why the `FileAccess` trait is exported, so I can implement it externally. If it turns out all right I can provide it upstream as well. I'm not assuming you want to add this dependency up front.

> While looking through this I was wondering what about this requires being in tokei itself, and not in its own crate. If there was a crate that allowed access to zips and tars in the same way that ignore's Walk struct does.

Because the current API mostly work with the filesystem stuff, any such implementation would currently have to provide unpacked filesystem paths. I'm trying to avoid that by putting all file accesses behind a trait that I can implement. The API I added does take an iterator. Any walker (including ignore's) that is an Iterator can in principle be used. If we want it to be fallible, as is the case for most archives, we might even want to make it `Iterator<Item = Result<F, Error>>`.

> Lastly for the use case of in memory collections, I believe you can use tokei currently to read in memory collections using parse_from_str, as a PathBuf can be entirely faked and it won't attempt to access the filesystem.

Sure, there are a couple of things you miss out on by using the lower-level APIs. Parallel processing as provided by `utils/fs.rs`, file extension and shebang detection. I'd providing a higher level API which would benefit from being included in future improvements and refactoring.

udoprog:
FTR, I'm currently working on building a PoC that uses this trait in combination with the `tar-rs`.

udoprog:
Do you plan on supporting anything like this?

For my project, first unpacking things into the filesystem in order to analyze them was a significant time sink.

XAMPPRocky:
@udoprog I would support unpacking tar files if the implementation was reasonable. I'd prefer to see this abstraction applied to a feature (such as tar archives) and see how that would impact the codebase before I would feel comfortable merging it in.

udoprog:
What do you mean @XAMPPRocky?

This change provides traits for hooking into existing file access with a default implementation for `&'a Path` to support what we do today.

An example tar implementation was done here (linked earlier in this thread):
https://github.com/udoprog/kernelstats/commit/813880587c32bb8b52a6a480c60d0c53dbd00861#diff-639fbc4ef05b315af92b4d836c31b023R116

Was there anything else you wanted?

XAMPPRocky:
I meant providing tar functionality in tokei itself.

PR #286, opened 2019-01-09
Don't use width larger than 79 by default

matklad:
closes https://github.com/Aaronepower/tokei/issues/277

I don't really understand the code, so this may be completely stupid, but why do we even try to make output wider than 79 characters by default?

I think it might make sense to try to make it *narrower* than 79 if the terminal itself is narrow, but if it is wide, it makes little sense to try to fill it completely.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -63,11 +63,17 @@ impl<'a> Cli<'a> {
             3: enable file level trace. Not recommended on multiple files")
         ).get_matches();
 
-        let columns = matches.value_of("columns").and_then(|s| s.parse().ok())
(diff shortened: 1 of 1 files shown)


Aaronepower:
Thank you for the PR! There is actually a reason for allowing variable terminal width. The main motivation is the `-f/--files` flag, which prints out individual file statistics with their file paths. File paths are automatically truncated if they would cause an overflow, so if you want to see longer file paths you need to have tokei expand to the maximum available width.

matklad:
That makes sense! So we should check if `-f` is present then?

matklad:
image

42 seconds ago :-)

PR #287, opened 2019-01-10
Adding "vhdl" file extension for VHDL files

xnorme:
VHDL files could be named with ".vhdl" extension as well as ".vhd"


Aaronepower:
Thank you for your PR, and congrats on your first contribution!

PR #294, opened 2019-01-21
Recongize ES modules as JavaScript

maxnordlund:
Node now supports [ES modules natively](https://nodejs.org/api/esm.html) using the .mjs extension.

I noticed this when running tokei on a repo written almost exclusively using ES modules and got way to few JS files.


Aaronepower:
Thank you for your pull request, and congrats on your first contribution!

Issue #249, opened 2018-08-25
Use box drawing characters?


Emerentius:
Skimming through the new comparison between tools, I noticed how all of them use the visually noisy hyphen `-----` lines. The [box drawing characters](https://en.wikipedia.org/wiki/Box-drawing_character) could be used to have clean, uninterrupted lines. These improve aesthetics and improve readability.
Of course, this requires unicode support. I wasn't sure if the choice of hyphens was deliberate to support legacy terminals.

Some comparisons:
```
-----------------------------------------------------------------------------------
 Language                Files        Lines         Code     Comments       Blanks
-----------------------------------------------------------------------------------
 Dockerfile                 14          251          219            6           26
 Markdown                    3          180          180            0            0
 Rust                      102        41073        28770         8630         3673
 Shell                       6          310          195           63           52
 TOML                        8          141          120            3           18
 XML                         1       134861       133755            0         1106
-----------------------------------------------------------------------------------
 Total                     134       176816       163239         8702         4875
-----------------------------------------------------------------------------------
```

```
───────────────────────────────────────────────────────────────────────────────────
 Language                Files        Lines         Code     Comments       Blanks
───────────────────────────────────────────────────────────────────────────────────
 Dockerfile                 14          251          219            6           26
 Markdown                    3          180          180            0            0
 Rust                      102        41073        28770         8630         3673
 Shell                       6          310          195           63           52
 TOML                        8          141          120            3           18
 XML                         1       134861       133755            0         1106
───────────────────────────────────────────────────────────────────────────────────
 Total                     134       176816       163239         8702         4875
───────────────────────────────────────────────────────────────────────────────────
```

Following LaTeX `booktabs` style:
```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
  Language                Files        Lines         Code     Comments       Blanks
───────────────────────────────────────────────────────────────────────────────────
 Dockerfile                 14          251          219            6           26
 Markdown                    3          180          180            0            0
 Rust                      102        41073        28770         8630         3673
 Shell                       6          310          195           63           52
 TOML                        8          141          120            3           18
 XML                         1       134861       133755            0         1106
───────────────────────────────────────────────────────────────────────────────────
 Total                     134       176816       163239         8702         4875
[...]

PR #307, opened 2019-02-28
Use box drawing characters if the charset is set to UTF8

Emerentius:
Implements #249

As described in the issue, this is a best efforts basis for POSIXy systems only. It checks if the codeset in `$LC_CTYPE` is set to UTF8 which is the same thing `tree` does. `LC_CTYPE` can be overriden by `LC_ALL` and falls back to `LANG` and so does this code.
While this code may be useful for someone else, I think it's not feature-complete enough for its own crate.

I've tested this under ubuntu 18.04, windows 7 (cmd, powershell and git bash). Ytain ran it on mac os where it also worked.

### DIFF
--- a/src/cli_utils.rs
+++ b/src/cli_utils.rs
@@ -66,13 +66,13 @@ pub fn print_header<W: Write>(sink: &mut W, row: &str, columns: usize)
     writeln!(sink, "{}", row)
 }
 
-pub fn print_results<'a, I, W>(sink: &mut W, row: &str, languages: I, list_files: bool)
+pub fn print_results<'a, I, W>(sink: &mut W, row: &str, columns: usize, languages: I, list_files: bool)
     -> io::Result<()>
     where I: Iterator<Item = (&'a LanguageType, &'a Language)>,
           W: Write,
 {
-    let path_len = row.len() - NO_LANG_ROW_LEN_NO_SPACES;
-    let lang_section_len = row.len();
+    let path_len = columns - NO_LANG_ROW_LEN_NO_SPACES;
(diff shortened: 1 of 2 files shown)


Aaronepower:
@Emerentius Thank you for your PR! Should `locale_requests_utf8` be in tokei however? I think it would fit better in a crate like `console` or `term` so that everyone in the Rust community can benefit from this kind of detection. I think this could be used for `console::Emoji` supporting more than just macOS.

src/cli_utils.rs
```diff
@@ -118,3 +118,48 @@ pub fn print_language<W>(sink: &mut W,
...
+pub fn locale_requests_utf8() -> bool {
+    // LC_ALL overrides LC_CTYPE which overrides LANG
+    // LC_CTYPE is the locale category used for the same purpose in the `tree` utility
+    let env_variables = ["LC_ALL", "LC_CTYPE", "LANG"];
+    let env = env_variables.iter()
+        .map(|&env| std::env::var(env))
+        .find(|env|
+            // filter out unset and empty variables
+            env != &Err(std::env::VarError::NotPresent)
+                && env != &Ok("".to_owned())
```
Aaronepower: This should be `&Ok(String::new())` instead of `&Ok("".to_owned())` for clarity.

XAMPPRocky:
```suggestion
                && env != &Ok(String::new())
```

XAMPPRocky:
I'm going to close for a lack of activity, however if someone wants to make a follow up to this PR that addresses my concerns above it more than welcome.

PR #319, opened 2019-03-21
languages.json tweak for improved Pascal accuracy

Akira13641:
This patch:  
  
- adds the `.pp` file extension, used heavily by Free Pascal in the set of libraries/packages distributed with it  
  
- removes the "mixed" `{` -> `*)` and `(*` -> `}` multiline comment styles, which are not actually valid in either of the two currently-widely-used Pascal compilers (Free Pascal and Delphi)  
  
- adds the `//` single-line comment style, which is supported by both FPC and Delphi and quite commonly used in a lot of code


Aaronepower:
@Akira13641 Thank you for your PR! It appears there are errors in the JSON and they need to be corrected before CI will pass.

Akira13641:
> @Akira13641 Thank you for your PR! It appears there are errors in the JSON and they need to be corrected before CI will pass. I believe the error is you did not remove the trailing comma in the `multi_line` property.  
  
Yikes, you're right! I was actually just looking and was like, "how could it fail CI? I didn't even change any real code!" Haha.

Akira13641:
Hmm, it looks like the CI failures now are only on some platforms, and unrelated to the JSON file. After fixing the trailing comma, Tokei does build fine for me specifically on x64 Windows (MSVC toolchain.)

Akira13641:
Seems that on Travis CI, all targets pass except i686-apple-darwin. On Appveyor, all the MSVC targets pass, but all the GNU ones fail. I did notice I should also add the `nested` flag set to true, though, so I'll do that now.

XAMPPRocky:
@Akira13641 Thank you again for your work, I've fixed some of the issues in CI, with travis passing I'm happy to merge this in. Congrats on your first contribution! :tada:

Akira13641:
Thanks!

PR #327, opened 2019-04-05
Add RPM specfile language

evitalis:
I was not 100% sure where to add the code to test said file but am happy to add it to this PR after some guidance if desired.

I work pretty regularly with RPMs and figured this would be helpful for myself and others.

### DIFF
--- a/tests/data/rpmspec.spec
+++ b/tests/data/rpmspec.spec
@@ -1,0 +1,42 @@
+# 43 lines 22 code 4 comments 16 blanks
+
+Name:           example
+Version:        0.0.1
+Release:        1%{?dist}
+Summary:        an example specfile
+
+Group:          
+URL:            
+Source0:        
+
+# test comments for requirements
+BuildRequires:  
+Requires:       
+
+%description
+
+
+%prep
+%setup -q
+
+
+# build the project
+%build
+%configure
+make build
+
+
+# install the files here
+%install
+make install
+
+
+%clean
+
+
+%files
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
@evitalis Thank you for your PR! You don't need to add any code you just a comment on the top of the file that says what stats the file should have. You can look at the other files in `tests/data` such as the [rust one](https://github.com/XAMPPRocky/tokei/blob/master/tests/data/rust.rs#L1) for an example.

Also if you could, could you change `"line_comment: ["#"]` to `"base": "hash"`? it does functionally the same thing but reduces the amount of code generated at compile time.

evitalis:
@XAMPPRocky I went ahead and added the counts. I also squashed my commits to make it a bit cleaner in the history.

XAMPPRocky:
@evitalis There's a small error in your JSON, `"base"` is expected to be a single string and not an array. :)

evitalis:
@XAMPPRocky Sorry for the back and forth hopefully this is in order now. I added a commit fixing the base definition.

tests/data/rpmspec.spec
```diff
@@ -0,0 +1,42 @@
+# 43 lines 22 code 4 comments 16 blanks
```
XAMPPRocky: This should be 42 lines.

evitalis: Fixed

XAMPPRocky:
Thank you again for your PR and congrats on your first contribution! 🎉

PR #331, opened 2019-05-06
Fix panic when columns < NO_LANG_HEADER_ROW_LEN

ocisly:
Thank you for making Tokei! This fixes the following panic by ensuring that `columns` is always wide enough to print the header row:

    $ cargo run -- --columns=66
    thread 'main' panicked at 'attempt to subtract with overflow',
    src/cli_utils.rs:64:17

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -47,16 +47,18 @@ fn main() -> Result<(), Box<Error>> {
         process::exit(0);
     }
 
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your pull request! Sorry the CI isn't working right now. This part of code isn't tested anyway 😅 So I'm happy to merge it in. Congrats on your first contribution! 🎉

PR #344, opened 2019-06-25
Added config so "/usr/bin/env ruby" works

### LINKED ISSUE #343: Support ruby scripts with /usr/bin/env shebang
It'd be useful for ruby projects to support `/usr/bin/env ruby` as a shebang line on ruby scripts - for example https://github.com/alphagov/verify-frontend/blob/master/bin/setup 

(PR incoming)

kornysietsma:
including a very simple test
fixes #343


XAMPPRocky:
Thank you for your PR!

PR #351, opened 2019-07-22
Add MoonScript to Languages

turbo:
A language that compiles to Lua, ref: https://moonscript.org/. Does not have multi-line comments.


XAMPPRocky:
Thank you for your PR, and congrats on your first contribution! 🎉🎉

PR #358, opened 2019-08-29
Add thrift language

dtolnay:
https://en.wikipedia.org/wiki/Apache_Thrift


XAMPPRocky:
Thank you for your PR!

PR #390, opened 2019-10-22
improve test data for python

rodrigocam:


XAMPPRocky:
Thank you for your issue! However I don't see what this improves, as it doesn't introduce new python syntax that might catch bugs. The updated count also isn't accurate.

PR #421, opened 2020-01-03
Add language support for Lex and Yacc

lipeng31:
The file extension 'y' is now associated to Yacc's source files and no
longer to Happy which I have no idea what language it is.

Signed-off-by: Li Peng <lipeng31@gmail.com>

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -708,12 +708,6 @@
                 "gvy"
             ]
         },
-        "Happy":{
-            "extensions":[
-                "y",
-                "ly"
-            ]
-        },
         "Handlebars":{
             "multi_line":[
                 ["<!--", "-->"],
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! However I would want to keep `Happy`, I would rather keep happy as it is and not have the `y` extension for yacc.

lipeng31:
Hi, thank you for your reply. Would you consider adding a command-line option to specify the language to disambiguate languages that share the same file extensions?

XAMPPRocky:
@lipeng31 Sorry for the late reply. I do plan to add this feature, however last time I attempted to implement it, I encountered a few hard limits with some dependencies and decided to leave it for the libraries to mature. I might try to revisit this soon.

PR #423, opened 2020-01-04
Fix License file names

Kixiron:


Kixiron:
Was informed that `LICENCE` is a European spelling, my apologies

PR #445, opened 2020-02-12
.pyw is a valid Python extension

mark-summerfield:
I added `pyw` since it is a valid Python extension.
I can only find two references to this fact in the official python docs (search for "pyw" in these):
https://www.python.org/dev/peps/pep-0397/
https://www.python.org/download/releases/2.4.1/notes/

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -1392,7 +1392,8 @@
                 "python3"
             ],
             "extensions":[
-                "py"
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! That's good enough for me. Congrats on your first contribution! 🎉

PR #457, opened 2020-03-04
Split out GNU style assembly

bwidawsk:
.asm is generally associated with MASM,NASM,YASM, while .s and .S are
associated with traditionally what is fed into the GNU assembler.

To support this, create two different assembly file types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,7 @@ Alex
 ASP
 ASP.NET
 Assembly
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR! Would you also be to update the list in the README.md?

XAMPPRocky:

PR #458, opened 2020-03-08
Optionally format numbers with thousands separator

### LINKED ISSUE #256: Optionally formatting numbers with a thousands separator
Would you be interested in including this functionality (hidden behind a flag of course). It's easier to read `49,341` than `49341`.

HarrisonMc555:
Closes #256.

Add a new option, `-n`/`--num-format` to specify whether the output should
separated by commas, dots, or nothing (plain).

This adds a new import, `num-format`, to perform the formatting. This adds a
total of three new dependencies: `arrayvec` version 0.4, `ndodrop`, and
`num-format`.

---

If we like the way this looks, I can add a configuration option. I would assume that most users would prefer to put their preferred value in the configuration option so they don't have to specify it every time.

### DIFF
--- a/src/cli_utils.rs
+++ b/src/cli_utils.rs
@@ -6,9 +6,10 @@ use std::{
 };
 
 use clap::crate_version;
+use num_format::{Format as NumFormat, ToFormattedString};
 
 use crate::input::Format;
-use tokei::{Language, LanguageType};
+use tokei::{Language, LanguageType, Stats};
 
 pub const FALLBACK_ROW_LEN: usize = 79;
 const NO_LANG_HEADER_ROW_LEN: usize = 67;
@@ -75,25 +76,32 @@ pub fn print_header<W: Write>(sink: &mut W, row: &str, columns: usize) -> io::Re
     writeln!(sink, "{}", row)
 }
 
-pub fn print_results<'a, I, W>(
+pub fn print_results<'a, I, W, F>(
     sink: &mut W,
     row: &str,
     languages: I,
     list_files: bool,
+    num_format: &F,
 ) -> io::Result<()>
 where
     I: Iterator<Item = (&'a LanguageType, &'a Language)>,
     W: Write,
+    F: NumFormat,
 {
     let path_len = row.len() - NO_LANG_ROW_LEN_NO_SPACES;
     let columns = row.len();
     for (name, language) in languages.filter(isnt_empty) {
-        print_language(sink, columns, language, name.name())?;
+        print_language(sink, columns, language, name.name(), num_format)?;
 
         if list_files {
             writeln!(sink, "{}", row)?;
             for stat in &language.stats {
-                writeln!(sink, "{:1$}", stat, path_len)?;
+                writeln!(
+                    sink,
+                    "{:1$}",
+                    FormattableStats::new(stat, num_format),
+                    path_len
+                )?;
             }
             writeln!(sink, "{}", row)?;
         }
@@ -106,14 +114,16 @@ pub fn isnt_empty(&(_, language): &(&LanguageType, &Language)) -> bool {
     !language.is_empty()
 }
 
-pub fn print_language<W>(
+pub fn print_language<W, F>(
     sink: &mut W,
     columns: usize,
     language: &Language,
     name: &str,
+    num_format: &F,
 ) -> io::Result<()>
 where
     W: Write,
+    F: NumFormat,
 {
     let mut lang_section_len = columns - NO_LANG_ROW_LEN;
     if language.inaccurate {
@@ -133,11 +143,11 @@ where
     writeln!(
         sink,
         "{:>6} {:>12} {:>12} {:>12} {:>12}",
-        language.stats.len(),
-        language.lines,
-        language.code,
-        language.comments,
-        language.blanks
+        language.stats.len().to_formatted_string(num_format),
+        language.lines.to_formatted_string(num_format),
+        language.code.to_formatted_string(num_format),
+        language.comments.to_formatted_string(num_format),
+        language.blanks.to_formatted_string(num_format)
     )
 }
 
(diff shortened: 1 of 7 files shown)


XAMPPRocky:
Thank you for your PR! I haven't found the time to properly review this yet, but would you be able to run the `benchmark.sh` file on these changes and paste the output here? I'd like to know if there is any regression in performance when the argument is not provided.

You need to have a released version of `tokei` and [`hyperfine`](https://github.com/sharkdp/hyperfine). I would also recommend cloning a large code repository like [`rust-lang/rust`](https://github.com/rust-lang/rust) to test with, then you can run the command like so.

```console
./benchmark.sh ./path/to/rust/repo
```

HarrisonMc555:
Ok here's my results. I'm not 100% sure how to interpret them, and I'm not feeling 100% confident in the results...it seems to be saying that my compiled version of `tokei` is 1.27 +/- 0.31 times faster than the installed version of `tokei`. That seems weird to me.

```text
$ ./benchmark.sh ~/git/rust
Tokei Benchmarking Tool
hyperfine 1.9.0
old tokei: tokei 10.1.2 compiled with serialization support: cbor, json, yaml
    Finished release [optimized] target(s) in 0.21s
Benchmark #1: target/release/tokei /Users/username/git/rust
  Time (mean ± σ):      2.689 s ±  0.143 s    [User: 10.738 s, System: 3.850 s]
  Range (min … max):    2.453 s …  2.891 s    10 runs

Benchmark #2: tokei /Users/username/git/rust
  Time (mean ± σ):      3.422 s ±  0.812 s    [User: 12.171 s, System: 4.026 s]
  Range (min … max):    2.822 s …  5.306 s    10 runs

  Warning: The first benchmarking run for this command was significantly slower than the rest (4.112 s). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.

Summary
  'target/release/tokei /Users/username/git/rust' ran
    1.27 ± 0.31 times faster than 'tokei /Users/username/git/rust'
```

XAMPPRocky:
@HarrisonMc555 Could you re run it. You can see there's a warning that the data is likely off, if you just re-run the benchmark, it will usually solve that.

HarrisonMc555:
I ran it twice more. The first gave the same warning about the first being slower, but the second didn't. Here it is:


```text
$ ./benchmark.sh ~/git/rust
Tokei Benchmarking Tool
hyperfine 1.9.0
old tokei: tokei 10.1.2 compiled with serialization support: cbor, json, yaml
    Finished release [optimized] target(s) in 0.18s
Benchmark #1: target/release/tokei /Users/username/git/rust
  Time (mean ± σ):      2.509 s ±  0.137 s    [User: 10.436 s, System: 3.731 s]
  Range (min … max):    2.391 s …  2.763 s    10 runs

Benchmark #2: tokei /Users/username/git/rust
  Time (mean ± σ):      2.790 s ±  0.130 s    [User: 11.990 s, System: 3.842 s]
  Range (min … max):    2.658 s …  3.050 s    10 runs

Summary
  'target/release/tokei /Users/username/git/rust' ran
    1.11 ± 0.08 times faster than 'tokei /Users/username/git/rust'
```

Just out of curiosity, is there an easy way to benchmark the current branch vs. `master`? I guess I could check out `master`, put it somewhere, make that a higher priority in my `PATH`, and then run this...

src/cli_utils.rs
```diff
@@ -75,25 +76,32 @@ pub fn print_header<W: Write>(sink: &mut W, row: &str, columns: usize) -> io::Re
...
+        print_language(sink, columns, language, name.name(), num_format)?;
 
         if list_files {
             writeln!(sink, "{}", row)?;
             for stat in &language.stats {
-                writeln!(sink, "{:1$}", stat, path_len)?;
+                writeln!(
+                    sink,
+                    "{:1$}",
+                    FormattableStats::new(stat, num_format),
```
XAMPPRocky: I'm not a big fan of the `FormattableStats` newtype, since it's only used for displaying the stats. I think what would an improvement would be to add a `display` method to `Stats` that takes `NumberFormatStyle` as an argument. This way the formatting API is also available to library authors.

```rust
pub fn display(&self, style: NumberFormatStyle) -> Result<String> {
}
```

I would also keep the `fmt::Display` implementation of `Stats` if possible since removing it will be a breaking change. So for now I'd prefer to keep that, and add a warning (`log::warn!`) to transition people to using the new method.

src/cli_utils.rs
```diff
@@ -151,3 +161,57 @@ where
...
+    num_format: &'a F,
+}
+
+impl<'a, F: NumFormat> FormattableStats<'a, F> {
+    pub fn new(stats: &'a Stats, num_format: &'a F) -> Self {
+        FormattableStats { stats, num_format }
+    }
+}
+
+macro_rules! display_stats {
```
XAMPPRocky: I know this is code that I wrote and not yours, but this really shouldn't be a macro since you should be able to easily write this as a function instead.

src/output.rs
```diff
@@ -0,0 +1,42 @@
+use num_format::{CustomFormat, Grouping};
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum NumberFormatStyle {
```
XAMPPRocky: Based on the feedback on `FormattableStats` this enum should be available in the library, and not just the binary, and should have `#[non_exhaustive]` so that we can add new formats without a major version bump.  What do you think about adding a `SystemLocale` variant that uses the `with-system-locale` feature in `num-format`?
```suggestion
#[non_exhaustive]
pub enum NumberFormatStyle {
```

XAMPPRocky:
> Just out of curiosity, is there an easy way to benchmark the current branch vs. master? I guess I could check out master, put it somewhere, make that a higher priority in my PATH, and then run this…

The easiest way without changes to `benchmark.sh` would be to use `cargo install --path .` on the `master` branch which will install master into your path, and then switch to your branch.

XAMPPRocky:
```
Summary
  'target/release/tokei /Users/username/git/rust' ran
    1.11 ± 0.08 times faster than 'tokei /Users/username/git/rust'
```
It's interesting that this shows these changes *at least* provide around 2% performance boost. This is probably because `std`'s formatting can be relatively slow. I will check this out further when we're close to merging.

src/cli_utils.rs
HarrisonMc555: I didn't think about the breaking change! That's a good point.

The only hard part about implementing `display` as you've suggested is that we don't have `fmt::Formatter`, which means we don't have a width. Is that a blocker?

src/cli_utils.rs
HarrisonMc555: I agree 😄 I was wondering why it was in a macro to begin with. I'll move it out.

src/output.rs
HarrisonMc555: Good idea! I like the suggestion for `#[non_exhaustive]`.

About adding a system locale feature, I just want to reference the description from `num_format` and make sure it's ok:

> Since this type requires several dependencies (especially on Windows), it is behind a feature flag. To use it, include num-format = { version = "0.4", features = ["with-system-locale"] } in your Cargo.toml. Additionally, on Windows (but only on Windows), using SystemLocale requires Clang 3.9 or higher.

From what I can tell (not 100% sure on this part), it looks like this would add the following dependencies:
 - `num-format-windows`
 - `widestring`
 - `winapi/winnls`

Figured I would check :smile:

src/cli_utils.rs
XAMPPRocky: You can add other arguments, what I said was a suggestion.

src/output.rs
XAMPPRocky: Yeah that's totally fine if it adds those dependencies

XAMPPRocky:
@HarrisonMc555 Are you still interested and/or able to work on this PR?

HarrisonMc555:
We actually just had our first child, so I've had less free time in the evenings. If someone else wants to pick this up that would be great 😄

NickHackman:
I'd be interested in picking this up after #535 :+1:

PR #470, opened 2020-03-22
Updated README.md to reflect support languages

SteveL-MSFT:
Started by noting that PowerShell wasn't listed in the README even though it is supported.  Then looked to see what else wasn't in the README based on languages.json.  Turns out a number of languages.  To make this easier to keep up-to-date, I simply grabbed all the language names from the json and replaced the ones in the README.  Some casing  has changed, but it seems that languages.json should be the source of truth which makes keeping the README up-to-date much simpler as well.


XAMPPRocky:

PR #478, opened 2020-03-25
[LICENSE-*] Consistent with others; [README.md] Added shield and same…

SamuelMarks:
… license text as rustlang/rustfix


XAMPPRocky:
Thank you for your PR, however I prefer to use British English and I'm not interested changing the spelling. Whether it's consistent with other repositories is not my concern.

PR #487, opened 2020-03-30
Check option in outer loop

pickfire:
Reduce the number of option checks by not distributing it at first.

Not sure if this helps improve the performance.
```
Tokei Benchmarking Tool
hyperfine 1.9.0
old tokei: tokei 11.0.0 compiled with serialization support: cbor, json, yaml
   Compiling tokei v11.0.0 (/home/ivan/src/pickfire/rs/tokei)
    Finished release [optimized] target(s) in 50.01s
Benchmark #1: target/release/tokei
  Time (mean ± σ):       5.4 ms ±   0.6 ms    [User: 5.7 ms, System: 4.4 ms]
  Range (min … max):     4.8 ms …  15.4 ms    510 runs

  Warning: Command took less than 5 ms to complete. Results might be inaccurate.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a
quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark #2: tokei
  Time (mean ± σ):       5.9 ms ±   1.2 ms    [User: 5.8 ms, System: 4.8 ms]
  Range (min … max):     4.8 ms …  12.8 ms    530 runs

  Warning: Command took less than 5 ms to complete. Results might be inaccurate.
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a
quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  'target/release/tokei ' ran
    1.08 ± 0.25 times faster than 'tokei '
```

### DIFF
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -1,6 +1,6 @@
 use std::{collections::BTreeMap, path::Path};
 
-use ignore::{overrides::OverrideBuilder, WalkBuilder, WalkState::Continue};
+use ignore::{overrides::OverrideBuilder, DirEntry, WalkBuilder, WalkState::Continue};
 
 use rayon::prelude::*;
 
@@ -87,39 +87,28 @@ pub fn get_all_files<A: AsRef<Path>>(
         })
     });
 
+    let rx_iter = rx
+        .into_iter()
+        .par_bridge()
+        .filter_map(|e| LanguageType::from_path(e.path(), &config).map(|l| (e, l)));
+
+    let process = |(entry, language): (DirEntry, LanguageType)| {
+        let result = language.parse(entry.into_path(), &config);
+        let mut lock = languages.lock();
+        let entry = lock.entry(language).or_insert_with(Language::new);
+        match result {
+            Ok(stats) => entry.add_stat(stats),
+            Err((error, path)) => {
+                entry.mark_inaccurate();
+                error!("Error reading {}:\n{}", path.display(), error);
+            }
+        }
+    };
+
     if let Some(types) = config.types.as_ref().map(|v| &**v) {
-        rx.into_iter()
-            .par_bridge()
-            .filter_map(|e| LanguageType::from_path(e.path(), &config).map(|l| (e, l)))
-            .filter(|(_, l)| types.contains(l))
-            .for_each(|(entry, language)| {
-                let result = language.parse(entry.into_path(), &config);
-                let mut lock = languages.lock();
-                let entry = lock.entry(language).or_insert_with(Language::new);
-                match result {
-                    Ok(stats) => entry.add_stat(stats),
-                    Err((error, path)) => {
-                        entry.mark_inaccurate();
-                        error!("Error reading {}:\n{}", path.display(), error);
-                    }
-                }
-            })
+        rx_iter.filter(|(_, l)| types.contains(l)).for_each(process)
     } else {
-        rx.into_iter()
-            .par_bridge()
-            .filter_map(|e| LanguageType::from_path(e.path(), &config).map(|l| (e, l)))
-            .for_each(|(entry, language)| {
-                let result = language.parse(entry.into_path(), &config);
-                let mut lock = languages.lock();
-                let entry = lock.entry(language).or_insert_with(Language::new);
-                match result {
-                    Ok(stats) => entry.add_stat(stats),
-                    Err((error, path)) => {
-                        entry.mark_inaccurate();
-                        error!("Error reading {}:\n{}", path.display(), error);
-                    }
-                }
-            })
+        rx_iter.for_each(process)
     }
 }
 
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -87,24 +87,40 @@ pub fn get_all_files<A: AsRef<Path>>(
         })
     });
 
-    let types: Option<&[LanguageType]> = config.types.as_ref().map(|v| &**v);
-
-    rx.into_iter()
-        .par_bridge()
-        .filter_map(|e| LanguageType::from_path(e.path(), &config).map(|l| (e, l)))
-        .filter(|(_, l)| types.map_or(true, |t| t.contains(l)))
-        .for_each(|(entry, language)| {
-            let result = language.parse(entry.into_path(), &config);
-            let mut lock = languages.lock();
-            let entry = lock.entry(language).or_insert_with(Language::new);
-            match result {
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! If you want to measure performance I would recommend using a code repository with a significant enough  lines of code to be able to accurately measure it. I would recommend trying [rust-lang/rust](https://github.com/rust-lang/rust) or the Unreal Engine (requires sign-up to download source).

pickfire:
@XAMPPRocky 3 runs, first run is fast maybe because I open chrome at the end.
```
> ./benchmark.sh ../rust
Tokei Benchmarking Tool
hyperfine 1.9.0
old tokei: tokei 11.0.0 compiled with serialization support: cbor, json, yaml
    Finished release [optimized] target(s) in 0.33s
Benchmark #1: target/release/tokei ../rust
  Time (mean ± σ):      3.454 s ±  1.024 s    [User: 8.769 s, System: 2.674 s]
  Range (min … max):    2.114 s …  5.274 s    10 runs

Benchmark #2: tokei ../rust
  Time (mean ± σ):      4.532 s ±  1.028 s    [User: 10.520 s, System: 3.397 s]
  Range (min … max):    3.451 s …  6.805 s    10 runs

Summary
  'target/release/tokei ../rust' ran
    1.31 ± 0.49 times faster than 'tokei ../rust'
> ./benchmark.sh ../rust
Tokei Benchmarking Tool
hyperfine 1.9.0
old tokei: tokei 11.0.0 compiled with serialization support: cbor, json, yaml
    Finished release [optimized] target(s) in 0.23s
Benchmark #1: target/release/tokei ../rust
  Time (mean ± σ):      4.870 s ±  0.832 s    [User: 12.317 s, System: 3.707 s]
  Range (min … max):    3.417 s …  5.633 s    10 runs

Benchmark #2: tokei ../rust
  Time (mean ± σ):      4.941 s ±  0.828 s    [User: 12.463 s, System: 3.823 s]
  Range (min … max):    3.357 s …  5.528 s    10 runs

Summary
  'target/release/tokei ../rust' ran
    1.01 ± 0.24 times faster than 'tokei ../rust'
> ./benchmark.sh ../rust
Tokei Benchmarking Tool
hyperfine 1.9.0
old tokei: tokei 11.0.0 compiled with serialization support: cbor, json, yaml
    Finished release [optimized] target(s) in 0.17s
Benchmark #1: target/release/tokei ../rust
  Time (mean ± σ):      4.787 s ±  0.789 s    [User: 12.072 s, System: 3.620 s]
  Range (min … max):    3.451 s …  5.472 s    10 runs

Benchmark #2: tokei ../rust
  Time (mean ± σ):      5.579 s ±  0.227 s    [User: 14.006 s, System: 4.266 s]
  Range (min … max):    5.139 s …  5.961 s    10 runs

Summary
  'target/release/tokei ../rust' ran
    1.17 ± 0.20 times faster than 'tokei ../rust'
```

src/utils/fs.rs
```diff
@@ -87,24 +87,40 @@ pub fn get_all_files<A: AsRef<Path>>(
         })
     });
 
-    let types: Option<&[LanguageType]> = config.types.as_ref().map(|v| &**v);
```
XAMPPRocky: Thank you for your PR, thinking on this I'm against this change. The current changes adds too much duplication to be worth a potential performance increase. However reviewing this I did figure out a way to refactor the condition that would result in less duplication and remove checking the option. I've included how it would look below. 

```rust
let types: &[LanguageType] = config.types.as_ref().map(|v| &**v).unwrap_or(&[]);
let types_is_empty = types.is_empty();

rx.into_iter()
// …
 .filter(|(_, l)| types_is_empty || types.contains(l))
```

pickfire: @XAMPPRocky but doesn't that still requires it to go through the `filter`? I don't think the compiler get to optimize that out, I also don't know how to check.

XAMPPRocky: Oh sorry, I actually didn't see that you didn't include that in the second branch. I'd still like to avoid having code so similar duplicated. I would accept an implementation that used a macro, though if it can be avoided, that would be preferred.

pickfire: I did try removing that line but the type have issues, and also I don't really want to use a `dyn`. What do you think of this?
```rust
let rx_iter = rx
    .into_iter()
    .par_bridge()
    .filter_map(|e| LanguageType::from_path(e.path(), &config).map(|l| (e, l)));

let process = |(entry, language): (DirEntry, LanguageType)| {
    let result = language.parse(entry.into_path(), &config);
    let mut lock = languages.lock();
    let entry = lock.entry(language).or_insert_with(Language::new);
    match result {
        Ok(stats) => entry.add_stat(stats),
        Err((error, path)) => {
            entry.mark_inaccurate();
            error!("Error reading {}:\n{}", path.display(), error);
        }
    }
};

if let Some(types) = config.types.as_ref().map(|v| &**v) {
    rx_iter.filter(|(_, l)| types.contains(l)).for_each(process)
} else {
    rx_iter.for_each(process)
}
```

XAMPPRocky: That looks good to me! I think that's about as DRY you can get without using dyn or macros.

pickfire: Okay, I added this. How should we test for regression?

XAMPPRocky: Feel free to.

pickfire:
@XAMPPRocky Does the CI checks for performance regression? If yes we could probably merge this right?
```
Tokei Benchmarking Tool
hyperfine 1.9.0
old tokei: tokei 11.0.0 compiled with serialization support: cbor, json, yaml
    Finished release [optimized] target(s) in 0.06s
Benchmark #1: target/release/tokei ../rust
  Time (mean ± σ):      6.115 s ±  0.817 s    [User: 14.755 s, System: 4.379 s]
  Range (min … max):    4.103 s …  6.789 s    10 runs

Benchmark #2: tokei ../rust
  Time (mean ± σ):      7.902 s ±  0.695 s    [User: 18.710 s, System: 5.725 s]
  Range (min … max):    6.706 s …  9.004 s    10 runs

Summary
  'target/release/tokei ../rust' ran
    1.29 ± 0.21 times faster than 'tokei ../rust'
```

XAMPPRocky:
CI doesn't check regressions, according to that it's at least 8% faster on your machine, that's good enough for me to merge and try it out locally, and if I can reproduce I'll put out a release.

Issue #514, opened 2020-05-06
Rust comment counting, counts number of comment blocks


NickHackman:
It appears that Tokei is counting the number of comment blocks rather than the number of comments.

**Version**: 11.1.0

## Test
```rust
use std::collections::{hash_map::Entry, HashMap};

use crate::stats::Stats;

/// Project Stats
///
/// Contains contributions for all contributors separated on language
#[derive(Debug, Default)]
pub struct ProjectStats {
    stats: HashMap<String, HashMap<&'static str, Stats>>,
}

impl ProjectStats {
    /// Construct a new ProjectStats
    pub fn new() -> Self {
        Self {
            stats: HashMap::new(),
        }
    }

    /// Get total contributions by name
    ///
    /// # Parameters
    ///
    /// name: S where S: AsRef<str> - name of the contributor
    ///
    /// # Returns
    ///
    /// Option<Stats>
    pub fn total_contribs_by_name<S: AsRef<str>>(&self, name: S) -> Option<Stats> {
        let contributions = self.stats.get(name.as_ref())?;
        Some(
            contributions
                .values()
                .fold(Stats::default(), |total, stats| total + stats),
        )
    }

    /// Get contributions by name broken down by language
    ///
    /// # Parameters
    ///
    /// name: S where S: AsRef<str> - name of the contributor
    ///
    /// # Returns
    ///
    /// Option<&HashMap<&'static str, Stats>>
    pub fn contribs_by_name<S: AsRef<str>>(
        &self,
        name: S,
    ) -> Option<&HashMap<&'static str, Stats>> {
        self.stats.get(name.as_ref())
    }

    /// Calculates the total amount of lines contributed to the Git Repository
    ///
    /// # Returns
    ///
    /// usize
    pub fn total_lines(&self) -> usize {
        let mut sum: usize = 0;
        for contributor in self.stats.values() {
            for stats in contributor.values() {
                sum += stats.lines;
            }
        }
        sum
    }

    /// Get an Iterator over all Contributors
    ///
    /// # Returns
    ///
    /// Iterator<Item = &String>
    pub fn contributors(&self) -> impl Iterator<Item = &String> {
        self.stats.keys()
    }

    /// Inserts into the nested HashMap<HashMap> structure
    ///
    /// # Parameters
    ///
    /// name: S where S: Into<String> - name of contributor
    ///
    /// lang: &'static str - language name
    ///
    /// stats: Stats
[...]

PR #517, opened 2020-05-09
Added Lifetimes edge case to Rust example

NickHackman:
Updated Rust test to have lifetimes, to catch this edge case.

#514


XAMPPRocky:
Thank you for your PR!

Issue #488, opened 2020-03-31
Vue: Javascript and CSS comments should be counted as comments instead of code


DragonI:
A Vue file may contain HTML, JS and CSS. Version 11 is only handling HTML comments. CSS and JS single line "//" and multi-line ["/*", "*/"] comments are not.  See below.

**Simple Vue component file**

```
<template>
  <!-- HTML comment -->
  <audio ref="audioFile" class="audio-deaf">
    <source
      :src="file"
      :type="type"
    >
    Your browser does not support the audio element.
  </audio>
</template>

<script>
/**
 * JSDoc comment
 */
export default {
  name: 'SlAudio',

  props: {
    /*
     * JS multi-line comment
     */
    file: {
      type: String,
      required: required
    },

    type: {
      type: String,
      required: false,
      default: 'audio/mp3'
    },

    show: {
      type: Boolean,
      required: false,
      default: false
    }
  },

  watch: {
    show: {
      handler (val) {
        if (val) {
          this.play()
        }
      }
    }
  },

  methods: {
    // single line comment
    play () {
      this.$emit('playing')
      this.$refs.audioFile.play()
    }
  }
}
</script>

<style lang="scss" scoped>
  // CSS comment
  .audio-deaf {
    margin-top: 20px;
  }
</style>

```

**v11 result**

```
 Language            Files        Lines         Code     Comments       Blanks
-------------------------------------------------------------------------------
 Vue                     1           66           58            1            7
-------------------------------------------------------------------------------
 Total                   1           66           58            1            7
```

XAMPPRocky:
Thank you for your issue! This is a known limitation of tokei where it does not correctly handle files that contain multiple languages. I don't know if this will ever be fixed, because frankly I haven't thought of a nice way to handle that corner case. I will try to think about it more, as enough people seem to be interested in this feature (languages in languages).

DragonI:
Hi.

"languages in languages" would be a major undertaking. What about keeping it simple? **No "languages in languages"**.

Instead, all HTML, JS and CSS code is counted as Vue code and comments are counted as comments. Not sure if it's as easy as this. But based on [languages.json](https://github.com/XAMPPRocky/tokei/blob/master/languages.json#L1167), add "line_comment" and modify "multi_line_comments" as below.

```
"line_comment": ["//"],
"multi_line_comments": [["<!--", "-->"],["/*", "*/"]],
```

New Vue config

```
"Vue": {
  "name": "Vue",
[...]

PR #534, opened 2020-05-18
Fixed HTML counting JS and CSS comments

NickHackman:
Previously all lines that were `CSS` or `JavaScript` comments were
counted as code, these are now properly counted as comments.

#488 applies to HTML as well :slightly_frowning_face:


XAMPPRocky:
Thank you for your PR!

PR #535, opened 2020-05-19
Verbatim strings

### LINKED ISSUE #330: C# line count is broken if verbatim string with backslash is present
**Describe the bug**
Count is broken when there is a line in C# file with verbatim string in it like this:
`private const string BasePath = @"a:\";`.

**To Reproduce**

1. Download file [SccTokeiFailure.zip](https://github.com/boyter/scc/files/3132587/SccTokeiFailure.zip)
2. Extract
3. `tokei`

**Actual behavior**
```
 C#                      1           20           17            0            3
 Total                   1           20           17            0            3
```

**Expected behavior**
```
 C#                      1           20           14            3            3
 Total                   1           20           14            3            3
```

**Tested on**

* Debian 9.7
* Windows 10

NickHackman:
Closes #330 

Adds support for verbatim strings in `Rust`, `CSharp`, `FSharp`, and `C++`. This should be easy enough to add for any other languages in the future. The only weird case for future contributors, here is that `Python` behaves correctly in the current implementation of Tokei and would behave incorrectly in the case of verbatim strings.

### Rust

The Rust implementation is fairly naive, ideally `languages.json` would have a regex `/r(#+)"/`, but the dependency on the `regex` crate would add a good chunk to the compile time and would hurt performance, as mentioned in your comment the most common types are implemented here.

### C++

I should've seen this coming, `C++` is incredibly annoying for verbatim strings, the syntax of [`"PREFIX()PREFIX"`](stroustrup.com/C++11FAQ.html#UD-literals) where `PREFIX` can basically be anything. This would ideally be a regular expression as well. An empty `PREFIX` appears to be the most common situation so that's currently what's implemented. It appears that `scc` doesn't work properly for this either as both bare `()`s and prefixed ones produce incorrect results.

Any advice/feedback is greatly appreciated :smile:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,11 +123,13 @@ The comment should use the syntax of the language you're testing.
 A good example of a test file is [`tests/data/rust.rs`].
 
 ```rust
-// 39 lines 32 code 2 comments 5 blanks
+// 41 lines 33 code 3 comments 5 blanks
 
 /* /**/ */
 fn main() {
-    let start = "/*";
+    let start = r##"/*\"
+\"##;
+    // comment
     loop {
         if x.len() >= 2 && x[0] == '*' && x[1] == '/' { // found the */
             break;
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,6 +65,24 @@ let x = /* There is a reason
     10;
 ```
 
+The `verbatim_quotes` property expects an array of strings, as some languages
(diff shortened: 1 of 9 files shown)


NickHackman:
CI pipeline timeout when trying to download a dependency, going to blank commit in order to re-trigger the pipeline.

NickHackman:
@XAMPPRocky should be ready for review now :smile:

XAMPPRocky:
Thank you for your PR!

PR #550, opened 2020-06-01
Specify the --sort option in the configuration files

### LINKED ISSUE #548: Specify the --sort option in the configuration file
It would be great if the configuration file was able to configure the default for the `--sort` cli option as it does for the `--type` option for example.

Hence, with a `.tokeirc` like this:
```toml
sort = "lines"
```
Calling `tokei` would be the same as calling `tokei --sort=lines`

Oxyriser:
Closes #548.

Allows to specify the --sort option in a configuration file.

~However as for now the sort option in a configuration file is case sensitive while the cli option is not.~

This is my first PR on this project and I'm a Rust beginner so I would be happy to receive some feedback.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -72,7 +72,11 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     print_header(&mut stdout, &row, columns)?;
 
-    if let Some(sort_category) = cli.sort {
+    let sort = cli
+        .sort
+        .or(config.sort);
+
+    if let Some(sort_category) = sort {
         for (_, ref mut language) in &mut languages {
             language.sort_by(sort_category)
         }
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -72,7 +72,11 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     print_header(&mut stdout, &row, columns)?;
 
-    if let Some(sort_category) = cli.sort {
+    let sort = cli
+        .sort
+        .or(config.sort);
+
+    if let Some(sort_category) = sort {
```
XAMPPRocky, line 79:
```suggestion
    if let Some(sort_category) = cli.sort.or(config.sort) {
```

XAMPPRocky:

PR #555, opened 2020-06-09
Fixed the language name Vim Script to Vim script

yuki-ycino:
The correct language name for vim is Vim script, not Vim Script.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -1232,7 +1232,7 @@
       "extensions": ["vcproj", "vcxproj"]
     },
     "VimScript": {
-      "name": "Vim Script",
+      "name": "Vim script",
       "line_comment": ["\\\""],
       "quotes": [["\\\"", "\\\""], ["'", "'"]],
       "extensions": ["vim"]


languages.json
```diff
@@ -1232,7 +1232,7 @@
       "extensions": ["vcproj", "vcxproj"]
     },
     "VimScript": {
-      "name": "Vim Script",
+      "name": "Vim script",
```
XAMPPRocky, line 1235:
```suggestion
      "name": "Vimscript",
```

Milly:
The correct name is "Vim script" in the help files of vim.
```sh
vim/runtime/doc> rg -w "Vim script" | wc -l
152
vim/runtime/doc> rg -w "Vimscript" | wc -l
0
```
Long ago "Vimscript" or "VimL" was also used, but it was removed because it was confusing.
At [patch 8.0.0360](https://github.com/vim/vim/commit/b544f3c81f1e6a50322855681ac266ffaa8e313c)

XAMPPRocky:
@Milly Thanks for the info! In that case seems fine to me.

XAMPPRocky:
@yuki-ycino Thank you for your PR, and congrats on your first contribution! 🎉

PR #573, opened 2020-06-23
Add saltstack and jinja support

AlexanderThaller:
Saltstack is just yaml with a different extention.

Jinja only supports {# #} for comments and they are multiline.

There can be jinja in saltstack files but havent figured out how to add support for that.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -595,6 +595,11 @@
       ],
       "extensions": ["js", "mjs"]
     },
+    "Jinja": {
+      "line_comment": ["#"],
+      "multi_line_comments": [["{#", "#}"]],
+      "extensions": ["jinja"]
+    },
     "Json": {
       "name": "JSON",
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! What does the Jinja inside the saltstack files look like? If I saw an example I would be able to tell you what would be required.

AlexanderThaller:
```

XAMPPRocky:
@AlexanderThaller Yeah in that case I would just also include the relevant Jinja syntax in the Saltstack definition.

AlexanderThaller:
Sorry I didn't really have a need for this anymore. If anybody wants to take over I'm fine with that. If not closing the PR is also fine.

XAMPPRocky:
I'm going to close it for now, if someone wants this feel free to open a new PR.

PR #588, opened 2020-07-10
Make --no-ignore imply all other --no-ignore- flags

### LINKED ISSUE #585: --no-ignore should imply parent and vcs
I think the current behavior is very confusing and unintuitive, as also evidenced by #452. Even if you do understand the behavior, having to specify all three flags is annoying. At least there should be an additional flag that turns off everything at once.

Looking at ripgrep, its `--no-ignore` implies `--no-ignore-dot`, `--no-ignore-exclude`, `--no-ignore-global`, `--no-ignore-parent`, `--no-ignore-vcs`, so I think tokei should follow that.

This would of course be a breaking change though. Not sure how you handle that, though I think this would be worth a breaking change :)

Livven:
Fixes #585.

### DIFF
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -51,7 +51,7 @@ pub fn get_all_files<A: AsRef<Path>>(
         .git_exclude(ignore_vcs)
         .git_global(ignore_vcs)
         .git_ignore(ignore_vcs)
-        .hidden(ignore && config.hidden.map(|b| !b).unwrap_or(true))
+        .hidden(config.hidden.map(|b| !b).unwrap_or(true))
         .ignore(ignore_dot)
         .parents(ignore && config.no_ignore_parent.map(|b| !b).unwrap_or(true));
 
--- a/src/utils/fs.rs
+++ b/src/utils/fs.rs
@@ -37,22 +37,23 @@ pub fn get_all_files<A: AsRef<Path>>(
         walker.overrides(overrides.build().expect("Excludes provided were invalid"));
     }
 
-    // Flip the booleans as ignore's semantics are the opposite of our options.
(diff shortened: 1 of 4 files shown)


XAMPPRocky:
Thank you for your PR! Would you be able to rebase this onto the latest master. Iy should make CI pass

Livven:
Done, but [one check](https://github.com/XAMPPRocky/tokei/pull/588/checks?check_run_id=924555028) was cancelled due to a timeout. Is there anything I can do to fix that? Haven't found a retry button myself :p

XAMPPRocky:
I've queued the job to run again it will hopefully pass this time. :)

src/utils/fs.rs
```diff
@@ -37,22 +37,23 @@ pub fn get_all_files<A: AsRef<Path>>(
...
-        .git_exclude(no_ignore_vcs)
-        .git_global(no_ignore_vcs)
-        .git_ignore(no_ignore_vcs)
-        .hidden(config.hidden.map(|b| !b).unwrap_or(true))
-        .ignore(config.no_ignore.map(|b| !b).unwrap_or(true))
-        .parents(config.no_ignore_parent.map(|b| !b).unwrap_or(true));
+        .git_exclude(ignore_vcs)
+        .git_global(ignore_vcs)
+        .git_ignore(ignore_vcs)
+        .hidden(ignore && config.hidden.map(|b| !b).unwrap_or(true))
```
Livven, line 54: Seems like `--no-ignore` would also imply hidden right now. Don't think that should be the case. This should be better:

```suggestion
        .hidden(config.hidden.map(|b| !b).unwrap_or(true))
```

I'll take a closer look and add a test tomorrow/later.

XAMPPRocky:
Thank again for your PR!

PR #591, opened 2020-07-17
feat: number formatted printing

### LINKED ISSUE #256: Optionally formatting numbers with a thousands separator

NickHackman:
Closes #256 and #458 

With the major release migrating #458 completely seemed like more work than starting from scratch as a result of the redesign of the printing system.

Adds command line argument to print numbers with commas, plain, or dots.
For example, dots would result in 1.324.

## Benchmark

```text
Tokei Benchmarking Tool
hyperfine 1.10.0old tokei: tokei 12.0.4 compiled with serialization support: json
    Finished release [optimized] target(s) in 0.07s
Benchmark #1: target/release/tokei ../rust
  Time (mean ± σ):     115.3 ms ±   2.4 ms    [User: 646.5 ms, System: 192.7 ms]
  Range (min … max):   111.6 ms … 121.1 ms    25 runs
 
Benchmark #2: tokei ../rust
  Time (mean ± σ):     116.3 ms ±   2.8 ms    [User: 657.2 ms, System: 190.3 ms]
  Range (min … max):   112.1 ms … 121.2 ms    25 runs
 
Summary  'target/release/tokei ../rust' ran    1.01 ± 0.03 times faster than 'tokei ../rust'
```

### DIFF
--- a/src/cli_utils.rs
+++ b/src/cli_utils.rs
@@ -9,7 +9,7 @@ use clap::crate_version;
 use num_format::ToFormattedString;
 
 use crate::input::Format;
-use tokei::{CodeStats, Language, LanguageType, Report};
+use tokei::{find_char_boundary, CodeStats, Language, LanguageType, Report};
 
 pub const FALLBACK_ROW_LEN: usize = 79;
 const NO_LANG_HEADER_ROW_LEN: usize = 67;
@@ -68,6 +68,14 @@ pub enum NumberFormatStyle {
     Commas,
     // 1.234
     Dots,
+    // 1_234
+    Underscores,
+}
+
+impl Default for NumberFormatStyle {
+    fn default() -> Self {
+        Self::Plain
+    }
 }
 
 impl FromStr for NumberFormatStyle {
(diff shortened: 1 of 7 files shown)


src/cli_utils.rs
```diff
@@ -331,3 +430,12 @@ impl<W: Write> Printer<W> {
...
 }
+
+fn find_char_boundary(s: &str, index: usize) -> usize {
+    for i in 0..4 {
+        if s.is_char_boundary(index + i) {
+            return index + i;
+        }
+    }
+    unreachable!();
+}
```
NickHackman, line 441: Copying and pasting this feels bad, but the options are either create methods inside of `Report` that are public and should really only be used by CLI version of Tokei or move over the required functionality to not making the library interface more confusing.

I'm not sure which is preferred here, both have their trade offs. I'm willing to move towards whichever is better for long term maintainability and library interface.

XAMPPRocky: I think I would be okay with a undocumented re-export.

XAMPPRocky:
Thank you for your PR!

shilangyu:
Hey are there any plans to release it? It seems to be not available in the newest version on crates.io

XAMPPRocky:
@shilangyu I’ll cut a release later this or next week.

PR #620, opened 2020-09-03
Add definition for DAML

axman6-da:
DAML is essentially a dialect of Haskell for the purposes of counting lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,7 @@ CSharp
 CShell
 Css
 D
+DAML
 Dart
 DeviceTree
 Dhall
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR! Would you also be able to add a test file under `tests/data`? You can look at the others as an example and there are also instructions for how to add the file in the `CONTRIBUTING.md`.

axman6-da:
Will do, just trying to find some code with an appropriate license!

XAMPPRocky:

PR #628, opened 2020-09-10
Add Matlab/GNU Octave, Scilab extension file

QrowPlissken:


XAMPPRocky:
Thank you for your PR! Would also be able to add a test file? How to do so should be documented in the contributing.md

languages.json
```diff
@@ -773,18 +778,11 @@
       "line_comment": ["#"],
       "extensions": ["nix"]
     },
-    "ObjectiveC": {
+    "ObjectiveCCpp": {
       "name": "Objective-C",
       "line_comment": ["//"],
       "multi_line_comments": [["/*", "*/"]],
       "quotes": [["\\\"", "\\\""]],
-      "extensions": ["m"]
```
XAMPPRocky, line 781: You shouldn’t remove another language. I would prefer if matlab didn’t include the `m` extension.

jedbrown: Reality is that `.m` is by far the most common extension for existing Matlab code. This is the logic used by `cloc` to disambiguate.
https://github.com/AlDanial/cloc/blob/master/Unix/cloc#L10870-L10993

PR #666, opened 2020-12-10
Add Terra support and Lua testing

seylerius:
This pull adds support for Terra, a strict-typed systems language that is metaprogrammable in Lua, as well as testing support for both Terra and Lua. I also took the liberty of adding more of the blockquote and block-comment variants Lua supports (you can have arbitrarily many `=`s between the square brackets to produce different levels of nesting), and moved the blockquotes to verbatim, as they _can_ include literal `"`s.


XAMPPRocky:
Thank you for your PR! It seems the JSON is invalid. Also are you sure it is correct for `[===[` to be considered a verbatim quote? Aren't they considered comments and not strings?

seylerius:
The difference between the block comment and string specifications for Lua
and Terra is simply the presence or absence of the initial dashes. All the
same rules about quantities of `=`s apply.

I'll run the JSON through `jq` real quick and see where the trouble is.
Silly me for missing it.

XAMPPRocky:
Thanks for the clarification. Then once this passes CI I would be happy to merge this in.

seylerius:
Okay, pushed those, let's see what we get. (Also, I just noticed I've got pull `666`, that's awesome).

seylerius:
I am n*no*t surprised about that failure. I hand-counted the lines, so it's gonna be a mixup in multi-line processing somewhere. I'll count 'em a fourth time by hand if that would be reassuring, though. (Hell, I'll go do that anyway just to reassure my own damn self.)

seylerius:
Okay, hand-counted again, same numbers: `91 lines 48 code 34 comments 9 blanks`. Definitely leaning toward the multi-line string and comment parsing being the source of this.

seylerius:
Can tokei spit out an annotated debug version that lists each line with why it was parsed that way?

XAMPPRocky:
Yes with `tokei -vvv`

XAMPPRocky:
Going to close this as inactive, if you'd still like to add support for this, feel free to open a new PR.

PR #678, opened 2020-12-23
Add --streaming flag

yijunyu:
Parallelize tokei counters and subsequent processing of individual files with additional options: 
```bash
tokei -f --cmd <shell_script> --folder <output_folder>
```
Here the `<shell_script>` is expected to accept three arguments: 
```bash
<shell_script> <filename> <language> <output_folder>
```
where the argument `<filename>` is the absolute path to the file found by `tokei`, the argument `<language>` is the programming language name identified by `tokei` according to the extension name of the file `<filename>`, and the argument `<output_folder>` is where the output folder to store the processed results.

# Why the pull request:
I work on Huawei's Trusted Programming project, which was recently asked to measure the code according to the programming languages, which is supported by `tokei` very well. 

We would like to do more with the tool so that one could:
* process the code files with a processor according to the programming language classified by `tokei`; 
* and do so in parallel especially when a large-scale code base is considered. 

However, the `Printer` impl struct has a mutable `writer` field, which blocks these subsequent processes for parallel processing. Hence, I've completed a working prototype for this by modifying the struct and move the `writer` field into function calls so that it is no longer `mutable`. As such, we could parallelize and speed up the subsequent processes.

If someone with more experience developing in `tokei` can give this an early look and send feedback, that would be really helpful. Apologies in advance for the current lack of tests and limited documentation. I'll be working on this full time for a while, so it will improve rapidly.

Thanks!


XAMPPRocky:
Thank you for your PR! Overall I'm in favour of adding this functionality though I think the implementation needs to be rewritten to better integrate with tokei.

The first thing is that `Printer` is not the best place for this functionality. While it's running in parallel, it's happening after tokei has already done the analysis, it would be more appropriate to hook into the existing parallel processing in the `process` closure in`utils::fs::get_all_files`.

Secondly is that tokei is also a library so I'd like to provide the ability to allow anyone to hook into it with their own functionality and then specialise in the CLI. How i think this would be implemented is that the `Config` has a `for_each_fn: Option<fn(LanguageType, Report)>` property (it needs to also be `#[serde(skip)]`) which is then called in the process closure.

Lastly, I don't think we should have `--cmd` or `--folder` flags, as these aren't really composable or cross platform compatible. Instead what I'd like to propose and get your feedback on, is having a `--streaming` flag which when combined with the `--output` flag will output the serialised `Report` object as each are analysed to stdout.

So when writing a script it would look something like this.

```bash
tokei src/ -o json --streaming | bash script.sh

# In script.sh
while read -r json
do
    echo "$json"
done
```

And the JSON would look like the following: (though obviously minimised).

```json
{
  "Rust": {
    "blanks": 7,
    "code": 26,
    "comments": 1,
    "reports": [
      {
        "stats": {
          "blanks": 7,
          "code": 26,
          "comments": 1,
          "blobs": {
            "Markdown": {
              "blanks": 30,
              "code": 0,
              "comments": 181,
              "blobs": {
                "Rust": {
                  "blanks": 5,
                  "code": 39,
                  "comments": 3,
                  "blobs": {}
                }
              }
            }
          }
        },
        "name": "src/lib.rs"
      }
    ],
    "children": {
      "Markdown": [
        {
          "stats": {
            "blanks": 30,
            "code": 0,
            "comments": 181,
            "blobs": {
              "Rust": {
                "blanks": 5,
                "code": 39,
                "comments": 3,
                "blobs": {}
              }
            }
          },
          "name": "src/lib.rs"
        }
      ]
    },
    "inaccurate": false
  }
}
```

yijunyu:
Thank you for the detailed review of the implementation. Indeed, we may distribute the proposed functionality into the three parts. 

Since `utils::fs::get_all_files` has already created the list of filenames in parallel, for our purpose, we don't need to touch that part and the `Printer` struct any more. 

Considering modifying the `output` and streaming it into bash script to parallel processing, the main thing is to produce the two arguments for the batch process. On the json output, I was able to use the `jq` utility to process the standard output produced by `-o json` as follows:
```bash
tokei src/ -o json | jq -r '{id: (keys | .[]), data: . } | select(.id != "Total") | { lang: .id, name: [(.data[.id] | .. | objects | with_entries(select(.key | contains("name"))) | select(. != {}) | .name)] } as $tuple | $tuple.name[] | . + " " + $tuple.lang'
```
, which could produce the list of files as follows:
```
src/lib.rs Rust
src/language/language_type.rs Rust
src/language/languages.rs Rust
src/language/mod.rs Rust
src/stats.rs Rust
src/sort.rs Rust
src/input.rs Rust
src/utils/ext.rs Rust
src/config.rs Rust
src/language/syntax.rs Rust
src/cli.rs Rust
src/language/language_type.tera.rs Rust
src/utils/macros.rs Rust
src/utils/mod.rs Rust
src/lib.rs Rust
src/language/language_type.rs Rust
src/language/languages.rs Rust
src/language/mod.rs Rust
src/stats.rs Rust
src/sort.rs Rust
src/main.rs Rust
src/cli_utils.rs Rust
src/input.rs Rust
src/utils/ext.rs Rust
src/config.rs Rust
src/utils/fs.rs Rust
src/language/syntax.rs Rust
src/cli.rs Rust
src/language/language_type.tera.rs Rust
```
These are essentially what we need for the batch processing shell script. When it comes to how to process this list of files in parallel.  `xargs` could handle that in the following piped command:
```bash
tokei src/ -o json | jq -r '{id: (keys | .[]), data: . } | select(.id != "Total") | { lang: .id, name: [(.data[.id] | .. | objects | with_entries(select(.key | contains("name"))) | select(. != {}) | .name)] } as $tuple | $tuple.name[] | . + " " + $tuple.lang' | xargs -n2 -P0 ./script.sh
```
However, as you suggested, this solution may not be as platform-independent as a Rust library could be because it depends on an external `jq` utility. Therefore, a `--streaming` option could still be useful to add as an additional function before the output. 

If you think it useful, I can implement the `--streaming` option for exactly this purpose. If you have a better idea, e.g., putting some JSON transformation into the Config, in case someone else have different needs, please let me know how to do that.

XAMPPRocky:
> If you think it useful, I can implement the --streaming option for exactly this purpose.

I do think it would be useful. Feel free to update this PR with the implementation, and post here if you have any questions about how to approach it or the codebase in general.

yijunyu:
*Update* The simplified option `--streaming` and its usage can be seen as follows: 
```bash
tokei --streaming -t=Rust,C,C++
```
It will select the list of files and their language types, one per line, for the downstream batch processing in the pipe. 
```
./build.rs Rust
./tests/accuracy.rs Rust
./src/lib.rs Rust
./src/utils/fs.rs Rust
./src/cli_utils.rs Rust
./src/language/language_type.tera.rs Rust
./src/language/languages.rs Rust
./src/language/syntax.rs Rust
./src/stats.rs Rust
./src/sort.rs Rust
./src/input.rs Rust
./src/main.rs Rust
./src/language/mod.rs Rust
./src/cli.rs Rust
./src/config.rs Rust
./src/utils/ext.rs Rust
./src/utils/macros.rs Rust
./src/utils/mod.rs Rust
./src/language/language_type.rs Rust
```

> Parallelize tokei counters and subsequent processing of individual files with additional options:
> 
> ```shell
> tokei -f --cmd <shell_script> --folder <output_folder>
> ```
> 
> Here the `<shell_script>` is expected to accept three arguments:
> 
> ```shell
> <shell_script> <filename> <language> <output_folder>
> ```
> 
> where the argument `<filename>` is the absolute path to the file found by `tokei`, the argument `<language>` is the programming language name identified by `tokei` according to the extension name of the file `<filename>`, and the argument `<output_folder>` is where the output folder to store the processed results.
> 
> # Why the pull request:
> I work on Huawei's Trusted Programming project, which was recently asked to measure the code according to the programming languages, which is supported by `tokei` very well.
> 
> We would like to do more with the tool so that one could:
> 
> * process the code files with a processor according to the programming language classified by `tokei`;
> * and do so in parallel especially when a large-scale code base is considered.
> 
> However, the `Printer` impl struct has a mutable `writer` field, which blocks these subsequent processes for parallel processing. Hence, I've completed a working prototype for this by modifying the struct and move the `writer` field into function calls so that it is no longer `mutable`. As such, we could parallelize and speed up the subsequent processes.
> 
> If someone with more experience developing in `tokei` can give this an early look and send feedback, that would be really helpful. Apologies in advance for the current lack of tests and limited documentation. I'll be working on this full time for a while, so it will improve rapidly.
> 
> Thanks!

src/config.rs
```diff
@@ -46,6 +46,9 @@ pub struct Config {
     pub types: Option<Vec<LanguageType>>,
     // /// A map of individual language configuration.
     // pub languages: Option<HashMap<LanguageType, LanguageConfig>>,
+    /// Whether to output only the paths for downstream batch processing 
+    /// *Default:* false
+    pub streaming: Option<bool>,
```
XAMPPRocky, line 51: `Config` shouldn't have a `streaming` option, because this option is only relevant for the CLI. `Config` should have a `for_each_fn: Option<fn (LanguageType, Stats)>` option, and then the `streaming` option in the CLI adds a closure to the `for_each_fn` option that then prints the result.

src/utils/fs.rs
```diff
@@ -91,15 +91,19 @@ pub fn get_all_files<A: AsRef<Path>>(
...
-            Ok(stats) => entry.add_report(stats),
-            Err((error, path)) => {
-                entry.mark_inaccurate();
-                error!("Error reading {}:\n{}", path.display(), error);
+        if ! config.streaming.unwrap_or_default() {
+            let result = language.parse(entry.into_path(), &config);
+            let mut lock = languages.lock();
+            let entry = lock.entry(language).or_insert_with(Language::new);
+            match result {
+                Ok(stats) => entry.add_report(stats),
```
XAMPPRocky, line 99: The `config.for_each_fn` should be called here before `entry.add_report(stats)`.

XAMPPRocky (commented):
Thank you for your new implementation. There's one change I'd like to make around how it behaves for libraries. Once that's done I'd be happy to merge this in.

src/config.rs
yijunyu: Adding the `pub for_each_fn: Option<fn (LanguageType, String)>` to the `Config` struct seems to cause the following compilation error:
```
the trait `serde::Deserialize<'_>` is not implemented for `fn(language::language_type::LanguageType, std::s
tring::String)`
```
Not sure how to add the `serde::Deserialize<'_>` trait yet, do you have any suggestion?

XAMPPRocky: Yes, sorry you need to add `#[serde(skip)]`.

src/utils/fs.rs
yijunyu: Thanks for the suggestion, fixed the current implementation accordingly.

src/config.rs
yijunyu: Please see if the updated code is OK, thanks!

src/config.rs
```diff
@@ -46,6 +46,11 @@ pub struct Config {
     pub types: Option<Vec<LanguageType>>,
     // /// A map of individual language configuration.
     // pub languages: Option<HashMap<LanguageType, LanguageConfig>>,
+    /// Whether to output only the paths for downstream batch processing 
+    /// *Default:* false
+    #[serde(skip)]
+    /// adds a closure for each function, e.g., print the result
+    pub for_each_fn: Option<fn (LanguageType, String)>,
```
XAMPPRocky, line 53: Sorry for the late review, I've been busy with work, but this should return a `Report`, and the report JSON with the language is what should be printed, not just the file path.

yijunyu: Will it be okay to have --streaming-json and --streaming as two separate options? Our batch processing would prefer a file at a line to simplify the bash downstreaming processes. If I generate JSON, it would require a json parser (an extra dependency on the jq tool).

XAMPPRocky: That sounds reasonable. I don't know about two separate options. I would instead change `--streaming` to accept an enum (see this example https://github.com/XAMPPRocky/tokei/blob/fd08c8b889f96986d3a05a17c58998ca062e9176/src/cli.rs#L135 ) with `Simple` and `Json` as its variants.

yijunyu: I made the change, please see if you would like it and whether the enum Streaming should be moved to another module or be kept with the cli.rs module.

XAMPPRocky: @yijunyu Thank you again for working on this. The streaming enum seems fine where it is. I still think this should be changed from `fn (LanguageType, String)` to `fn (LanguageType, Report)`. Could we also change the simple format to include stats, I think people would be interested in that information. I think the simple format should look as follows.

```
{language} {path} {lines} {code} {comments} {blanks}

Rust src/main.rs 50 25 15 10
```

src/cli.rs
```diff
@@ -229,6 +261,13 @@ impl<'a> Cli<'a> {
             config.no_ignore_vcs
         };
 
+        config.for_each_fn = match self.streaming {
+            Some(Streaming::Json) => Some(|l: LanguageType, e| println!("{{language: '{}', stats: '{}'}},", l.name(), serde_json::json!(e))),
```
XAMPPRocky, line 265:
```suggestion
            Some(Streaming::Json) => Some(|l: LanguageType, e| println!("{}", serde_json::json!({ "language": l.name(), "stats": e  }))),
```

src/config.rs
yijunyu: Got this implemented on `fn <LanguageType, Report>`, easier than I initially thought, thanks to the predefined formatter.

src/cli.rs
yijunyu: For the "simple" output, it prints a row per file. If you think it is useful to include a header line, I can add it with a leading "#" so that the batch process could ignore it. Let me know.

yijunyu: Not sure if the default `--streaming=none` should be kept in the command line options, if you think it does no harm, I will consider the implementation being completed :-)

XAMPPRocky: Well I don't think there needs to be a none option. None is implied by it not being present in the command.

yijunyu: Done !

yijunyu: If OK, is it possible to merge this into the trunk, or if you would like to add any more test cases?

PR #679, opened 2020-12-23
Add support for pug

### LINKED ISSUE #671: Add support for .pug files
[Pug.js](https://pugjs.org/) is a template engine language that renders HTML. A pug template can be written in `.pug` files.

***

I would happily make a PR myself, but I'm unsure of what to put for the `"quotes"` key. The issue is that quotes (e.g. `"`) are sometimes quotes for strings and sometimes literal quotes.

## Example

```pug
// example.pug
doctype html
html
  head
  body
    h1 "Hello, #{ "World" }!"
```
Becomes (indentation added for readability)
```html
<!doctype html>
<html>
  <head></head>
  <body>
    <h1>"Hello, World!"</h1>
  </body>
</html>
```

***

As you can see, the `"` outside of `#{ }` is a literal quote, while the `"` inside `#{ }` are quotes around a string. Per my understanding, this is the exact opposite behavior of the `"verbatim_quotes"` -- `"` is a verbatim quote *unless* it's wrapped by `#{ }`.

There can be any number of spaces inside `#{ }`, so I don't know if `"quotes"` can just be set to `[["#{\\\"", "\\\"}"], etc.]`.

spenserblack:
Closes #671

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -935,6 +935,15 @@
       "line_comment": ["//"],
       "extensions": ["proto"]
     },
+    "Pug" : {
+      "line_comment": ["//", "//-"],
+      "quotes": [
+        ["#{\\\"", "\\\"}"],
+        ["#{'", "'}"],
+        ["#{`", "`}"]
+      ],
+      "extensions": ["pug"]
+    },
     "PureScript": {
       "nested": true,
       "line_comment": ["--"],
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR!

Issue #684, opened 2020-12-29
Total of files displayed as zero


Byron:
Starting from 34810da843ebb21fb984de0bd8832a3bfd625315 (as found via `git bisect`) `tokei` doesn't seem to compute the total amount of files counted anymore.

Personally I found this informative so I wonder if there was a reason to remove the count and I am missing something, or if this could be a regression.

Thanks a lot.

XAMPPRocky:
Thank you for your issue! That's not intentional, just an unfortunate side effect of the change. I'll publish a fix later today.

PR #685, opened 2020-12-29
Fix for #684

Byron:
This is the simplest way to restore the original tokei file counting
functionality, even though it might not be the most efficient thanks
to various clones done within `summarise()`.

I also tried to count `children`, but that doesn't yield the correct result.

### DIFF
--- a/src/language/languages.rs
+++ b/src/language/languages.rs
@@ -106,6 +106,7 @@ impl Languages {
             total.code += language.code;
             total.inaccurate |= language.inaccurate;
             total.children.insert(*ty, language.reports.clone());
+            total.reports.extend(language.summarise().reports);
         }
         total
     }


XAMPPRocky:
Thank you for your PR! However I don't want to add that back, as it pollutes the JSON output for Total with redundant data. I think the proper fix belongs in `src/cli.rs`, and changing `print_total` to not call `print_language`, and instead having it's own version that calls `language.children.iter().sum(|c| c.reports.len())`. If you want to work on that feel free to update this PR and if you have more questions feel free to ask.

https://github.com/XAMPPRocky/tokei/blob/master/src/cli_utils.rs#L174

XAMPPRocky:
Hmm, for some reason my comment disappeared. If you still want help let know.

Byron:
Thanks a lot for your reply! I had a feeling this wouldn't be so easy (even though the fixed works for my immediate needs).

Thus I believe it's best to close this issue and leave the fix to those who know the code base better.

Thanks for making and maintaining `tokei`, it's a tool I use a lot to reassert myself of the work done by the end of a day and of course to get an idea of the composition of projects I just cloned. In case you didn't know.

PS: In case you didn't know [`onefetch`](https://lib.rs/crates/onefetch), it uses tokei to provide a nice overview, too.

XAMPPRocky:
@Byron The fix is still pretty simple, and PR is still welcome. It's just replacing the `print_language` call in `print_total` line with a copy of the `print_language` function and replacing `language.reports.len()` with `language.children.iter().map(|c| c.reports.len().sum()`.

- `print_language`: https://github.com/XAMPPRocky/tokei/blob/master/src/cli_utils.rs#L431
- `print_total`: https://github.com/XAMPPRocky/tokei/blob/master/src/cli_utils.rs#L174

PR #687, opened 2020-12-29
Fix all feasible clippy issues

Byron:
And the ones not deemed feasible were officialy allowed so folks running
`cargo clippy` should see no warnings.

As `clippy` isn't always perfect and sometimes is a matter of taste, please feel free to make the modifications right away in this PR or close outright in case it's not desired.

I would have wanted to enable clippy in CI but didn't want to risk breakage or overstep 😅.

### DIFF
--- a/src/cli_utils.rs
+++ b/src/cli_utils.rs
@@ -242,7 +242,7 @@ impl<W: Write> Printer<W> {
         Ok(())
     }
 
-    fn print_code_stats<'a, 'b>(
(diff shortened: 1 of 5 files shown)


XAMPPRocky:
Thank you for your PR!

> I would have wanted to enable clippy in CI but didn't want to risk breakage or overstep

Yeah I don't find putting clippy in CI to be incredibly valuable, I prefer to just accept PRs that fix them.

PR #690, opened 2021-01-04
Implement compact output feature (#605)

### LINKED ISSUE #605: Feature: --comact option for hiding another language inside of one.
As discussed [here](https://github.com/XAMPPRocky/tokei/issues/602), at the moment there's no way to disable showing languages inside language, e.g. Markdown inside Rust, so it would be useful to have an option to disable nested languages, for showing sorted lines of code ignoring nested languages.

Also I'd like to implement this myself, if nobody minds.

alexmaco:
Fixes #605 

Help text is a good bikeshed topic.

Although it may make sense to avoid searching for embedded languages entirely to maybe gain some performance when running with `--compact`, I left it as an output flag at first to avoid potential counting differences around edge cases.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -20,6 +20,7 @@ pub struct Cli<'a> {
     pub print_languages: bool,
     pub sort: Option<Sort>,
     pub types: Option<Vec<LanguageType>>,
+    pub compact: bool,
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR!

PR #707, opened 2021-01-29
added support for kvlang - https://kivy.org/doc/stable/guide/lang.html

rrrodzilla:
Hi there,

Thanks for the great tool.  Python developers who write GUIs using the Kivy framework define much of the UI in .kv language files (.kv).  I've added the language definition for kv lang to the languages.json file and added a test file that passes just fine.  
You can read more about kv lang here: https://kivy.org/doc/stable/guide/lang.html

Thanks again,

Roland


XAMPPRocky:

PR #717, opened 2021-02-18
Change display name of Lisp to Common Lisp

Janfel:
The term “Lisp” refers to a family of languages including Common Lisp, Emacs Lisp, Clojure, Arc, LFE and the entire Scheme family. While Common Lisp is an important member, it is neither the first nor the only one. It is incorrect to refer to Common Lisp as just Lisp, especially as `tokei` also supports other Lisp dialects like Scheme and Emacs Lisp.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -731,6 +731,7 @@
       "extensions": ["lds"]
     },
     "Lisp": {
+      "name": "Common Lisp",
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! Sorry for the late review.

Issue #333, opened 2019-05-09
FR: List of file paths to ignore


turbo:
I'm aware that tokei has an exclude option, but in my case, it's scanning hundreds of thousands of files. A lot of them have to be ignored on a single file basis (no pattern, not ignoring a directory). Since this will exceed the maximum command line length, I'd like to be able to give the exclude option a file with a list of paths to ignore.

In general, I'd like to supply configuration to tokei via a file instead of command line parameters, but that's not what this request is about (though it would be an elegant way to solve this - TOML or something).

XAMPPRocky:
@turbo Tokei does respect any `.tokeignore` files in the starting directory, this file uses the same syntax as `gitignore`. You can see an example of this in the tokei repository itself.

PR #721, opened 2021-02-24
README.md: describe .tokeignore file

mykter:
Should help pre-empt questions like #333

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,9 @@ the same semantics as `.gitignore`.
 $ tokei ./foo --exclude *.rs
 ```
 
+Paths to exclude can also be listed in a `.tokeignore` file, using the same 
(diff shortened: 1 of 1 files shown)


XAMPPRocky:

mykter:
You're welcome, thanks for the awesome tool!
I was running cloc on my project and it ran so slowly I suddenly thought "is there a ripgrep but for cloc?". Turns out there is and it totally beat my expectations 😄

XAMPPRocky:
> I was running cloc on my project and it ran so slowly I suddenly thought "is there a ripgrep but for cloc?". Turns out there is and it totally beat my expectations

Indeed, tokei in fact uses a lot of the same dependencies from ripgrep. 🙂

Issue #725, opened 2021-02-25
Markdown: double-counting of a comment


mykter:
Given a markdown file:
```markdown
```r    ```
```

tokei reports:
```
$ tokei test.md
===============================================================================
 Language            Files        Lines         Code     Comments       Blanks
===============================================================================
 Markdown                1            2            0            2            0
 |- R                    1            0            0            0            0
 (Total)                              2            0            2            0
===============================================================================
 Total                   1            2            0            2            0
===============================================================================
```

I found this through fuzzing, I'll raise a PR with the fuzz harness shortly.

PR #726, opened 2021-02-25
Fuzzing

mykter:
Add a cargo-fuzz harness to make it easy to fuzz test the file parser for robustness and accuracy.

See fuzz/README.md for details.

With only a few seconds run time I found #725 - I suspect there are many more, if this kind of consistency bug is of interest.

When looking for panics only, after a few hours and 80 million executions, it found #727. There may be more to find after this one is fixed.

I am a novice Rust programmer - apologies in advance for blunders; any suggestions for better ways of doing things are welcome.

The only change to the existing tokei code is to make each LanguageType derive from [Arbitrary](https://docs.rs/arbitrary/1.0.0/arbitrary/trait.Arbitrary.html). This makes it really easy to write a clean fuzzing harness. If it's desirable to avoid including this in release builds, I think we could conditionally compile it using [cfg(fuzzing)](https://rust-fuzz.github.io/book/cargo-fuzz/guide.html#cfgfuzzing)?


XAMPPRocky:
Thank you for your PR, and related issues! Overall this looks great! I'll give a more in depth review some time soon.

XAMPPRocky:
Thank you for your PR! I'm going to merge this in for now, as it looks mostly fine to me. 

>  I suspect there are many more, if this kind of consistency bug is of interest.

Yes, I would like tokei to be consistent.

mykter:
Great, I'll run the fuzzer again when the existing issues are fixed.

PR #730, opened 2021-02-26
Cache embedding regexes

### LINKED ISSUE #616: Infinite loop on some encodings
I tried running tokei (12.0.4) on a directory with a bunch of source in it and it went into an infinite loop. I tried to track it down a bit and it looks like it's related to Firefox encoding tests.

Here are some files it failed on (you can download the raw file from these pages to test):

* https://searchfox.org/mozilla-central/source/testing/web-platform/tests/encoding/legacy-mb-japanese/shift_jis/sjis_chars-csshiftjis.html
* https://searchfox.org/mozilla-central/source/testing/web-platform/tests/encoding/legacy-mb-korean/euc-kr/euckr_chars-cseuckr.html
* https://searchfox.org/mozilla-central/source/testing/web-platform/tests/encoding/legacy-mb-tchinese/big5/big5_chars-big5-hkscs.html

There are also others. I suggest downloading the Firefox code and running tokei on it to verify.

alexmaco:
Fixes #616 

This runs the embedding regexes once per line, and caches their results. The state machine then only tries to look up results by the index.

This also adds a test suite dedicated to embedding, containing only one test file, lifted from the firefox 85 codebase. This is done since adjusting the existing html test to reproduce some problems seemed more complicated.

As a side effect of these changes, counting has become more accurate around embedding (e.g. tokei preciously did not detect javascript at all in the included embedding test), so the precise behavior is expected to differ from the current master.

There are a few more miscounting issues lurking in there, but they seem to be preexisting. Eventually the last commit that simplifies the embedding test should be reverted.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -69,7 +69,18 @@ fn generate_tests(out_dir: &OsStr) -> Result<(), Box<dyn error::Error>> {
     const INITIAL_BUFFER_SIZE: usize = 989 * 130;
     let mut string = String::with_capacity(INITIAL_BUFFER_SIZE);
 
-    let walker = Walk::new("./tests/data/").filter(|p| match p {
+    generate_tests_batch("./tests/data", None, &mut string)?;
+    generate_tests_batch("./tests/embedding", Some("embedding"), &mut string)?;
+
+    Ok(fs::write(Path::new(&out_dir).join("tests.rs"), string)?)
+}
+
+fn generate_tests_batch(
+    src_dir: &str,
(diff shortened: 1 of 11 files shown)


XAMPPRocky:
Thank you for your PR! I will try to review this sometime soon.

XAMPPRocky:
Thank you for your PR! I think there's a small amount of cleanup that might be needed in the new module, but I'm still going to merge this, and we can iterate further with future PRs.

Issue #660, opened 2020-11-19
Incorrect help message for --languages?


spenserblack:
According to `tokei --help`:

>```
>FLAGS:
>    ...
>    -l, --languages           Prints out supported languages and their extensions.
>```

Unless I'm missing something, [tokei prints *only* the supported language](https://github.com/XAMPPRocky/tokei/blob/v12.0.4/src/cli.rs#L151), *not* the extension.

PR #745, opened 2021-03-17
Add extensions output for --languages options

maru143:
Hi,

As written in issue #660, there is an inconsistency between the help message about `--languages` and actual output.
So I tried to fix them by adding language extensions to the output (because it seems more useful for me).

But I wonder how the output should be arranged. Currently I'm using whitespace as a separator, but is it preferred to be printed in other styles like comma-separated?

### DIFF
--- a/src/language/language_type.tera.rs
+++ b/src/language/language_type.tera.rs
@@ -54,8 +54,13 @@ impl LanguageType {
     }
 
     /// Provides every variant in a Vec
-    pub fn list() -> &'static [Self] {
-        &[{% for key, _ in languages %}{{key}}, {%- endfor %}]
+    pub fn list() -> &'static [(Self, &'static [&'static str])] {
+        &[{% for key, val in languages -%}
+            ({{key}}, 
+            {% if val.extensions %} &[{% for extension in val.extensions %}"{{extension}}", {% endfor %}],
+            {% else %} &[],
+            {% endif %}),
+        {% endfor %}]
     }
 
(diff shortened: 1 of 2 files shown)


src/language/language_type.tera.rs
```diff
@@ -54,8 +54,13 @@ impl LanguageType {
     }
 
     /// Provides every variant in a Vec
-    pub fn list() -> &'static [Self] {
-        &[{% for key, _ in languages %}{{key}}, {%- endfor %}]
+    pub fn list() -> &'static [(Self, &'static [&'static str])] {
```
spenserblack, line 57: Since this changes the return type of `LanguageType::list`, it is a breaking change for users that currently use tokei as a library. It might be better to create a new `fn` instead.

XAMPPRocky:
Thank you for your PR! Sorry for the late review.

> But I wonder how the output should be arranged. Currently I'm using whitespace as a separator, but is it preferred to be printed in other styles like comma-separated?

I think my preference would be for `<Language> (<extension>, <extension>)` E.g. `JavaScript (js, es, javascript)` .

src/language/language_type.tera.rs
XAMPPRocky: The breaking change is fine, I intend to release a new major version.

maru143: Sorry for my late reply, and thank both of you for your review!

XAMPPRocky:
Thank you for your PR!

Issue #769, opened 2021-05-18
Cargo install throwing issue during installations


srijan-sivakumar:
The cargo command when run with `cargo install --git https://github.com/XAMPPRocky/tokei.git`
fails as there are multiple packages with binaries inside the project. For a particular package,
we'd have to add the name of the package, hence adding `tokei`

PR #770, opened 2021-05-18
Cargo install throwing issue during installations

srijan-sivakumar:
The cargo command when run with cargo install --git https://github.com/XAMPPRocky/tokei.git
fails as there are multiple packages with binaries inside the project. For a particular package,
we'd have to add the name of the package, hence adding tokei

Fixes: #769


XAMPPRocky:

srijan-sivakumar:
@XAMPPRocky , the CI seems to be failing in case of Windows and macos.. not sure why a readme change is causing that effect

PR #778, opened 2021-06-01
Improve readme

throwaway1037:
- "Linux" -> GNU+Linux
- Clarify different systems
- Give examples of distros


XAMPPRocky:
Thank you for your PR! However I’m not sure I agree that it is an improvement. I think it places too much emphasis on what Unix derivative it is, which I don’t think is important here. I want to keep macOS and Windows at the same heading level as Linux+others, as those platforms are where most people are.

XAMPPRocky:
Going to close this as I don't see the value in the change, but your contribution is appreciated.  🙂

PR #781, opened 2021-06-08
Add menhir support

pjmkrpg:
`.mll` and `.mly` files allow more comment styles than in Ocaml


XAMPPRocky:
Thank you for your PR! Sorry for the late reply, but it seems the count is slightly off, would you be able to address that?

XAMPPRocky:

PR #801, opened 2021-07-17
Fix clippy warnings

lespea:
There's also a deprecated function that I didn't want to just change but I'll point it out... in `language/synatx` line 94 you use the `byte_classes` method on the `AhoCorasickBuilder` but apparently [that either doesn't do anything anymore or won't shortly](https://github.com/BurntSushi/aho-corasick/issues/57).


XAMPPRocky:

PR #803, opened 2021-07-24
Add support for Raku

CIAvash:
Perl6 has been renamed to [Raku](https://raku.org/), so I removed Perl6 and added Raku. Also the old definition can result in incorrect calculations. I added more data, but even that is not enough, because Raku accepts many brackets and has more POD features, but it should cover most use cases.

Right now tests fail, because in Raku, a single comment starts with `#`, but POD declarations and multi-line comments start with `#` as well, with the difference in the character(s) that comes after `#`. For example, a multi-line comment can be written as ``#`(a multi-line comment)``, and a POD declaration can be written as `#| Some declaration` (POD declarations can be multi-line too).

So for that reason I think tokei is matching everything that starts with `#` as comment, and therefore, multi-line comments and POD declarations are ignored.


XAMPPRocky:
Thank you for your PR! Sorry for the late reply, it seems like the tests are failing, tokei reports only 36 lines of code, while your comment states there's 38. Would you be able to investigate why that is the case, you can use `tokei -vvv` to see individual line analysis.

CIAvash:
OK, the `tokei -vvv` helped. Because there is no `single_line_doc`, I used newline in `doc_quotes` which doesn't do anything I guess(end of line would be nice too). Also some doc lines need double newlines or end of line:
```json
"doc_quotes": [
        ["#|", "\n"],
        ["#=", "\n"],
        ...
        ["=item", "\n\n"],
        ["=head", "\n\n"],
]
```

```
[2022-08-14T11:45:11Z TRACE tokei::language::language_type] #| Fibonacci with multiple dispatch
[2022-08-14T11:45:11Z TRACE tokei::language::syntax] Start Doc "#|"
[2022-08-14T11:45:11Z TRACE tokei::language::syntax] End "\n"
[2022-08-14T11:45:11Z TRACE tokei::language::language_type] #| Fibonacci with multiple dispatch
[2022-08-14T11:45:11Z TRACE tokei::language::language_type] Comment No.1
[2022-08-14T11:45:11Z TRACE tokei::language::language_type] Was the Comment stack empty?: true
```

```
[2022-08-14T11:45:11Z TRACE tokei::language::language_type] #|{
[2022-08-14T11:45:11Z TRACE tokei::language::syntax] Start Doc "#|"
[2022-08-14T11:45:11Z TRACE tokei::language::syntax] End "\n"
[2022-08-14T11:45:11Z TRACE tokei::language::language_type] #|{
[2022-08-14T11:45:11Z TRACE tokei::language::language_type] Comment No.2
[2022-08-14T11:45:11Z TRACE tokei::language::language_type] Was the Comment stack empty?: true
```

So right now I don't know a way to detect doc lines. A workaround could be treating some docs(the ones that start with `#`) as comments and ignore the others.

XAMPPRocky:
> So right now I don't know a way to detect doc lines. A workaround could be treating some docs(the ones that start with #) as comments and ignore the others.

For now could you open a seperate issue to handle that case? Then we can merge this without those (and you could change the single line ones to single line comments for now), tokei is never perfectly accurate, there's always corner cases it won't catch.

CIAvash:
Sorry I made a mess of commit messages for a moment, it should be OK now.

XAMPPRocky:

PR #827, opened 2021-09-15
README: FreeBSD and NetBSD are not Linux

alyssais:
FreeBSD and NetBSD are both independent operating systems, with their own kernels, libcs, rust triples, etc.  They are not Linux distributions.  So move them out of the Linux section into their own sections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,14 +94,20 @@ cargo install tokei
 conda install -c conda-forge tokei
 # Fedora
 sudo dnf install tokei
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! While what you say is true, I don't see the value in having them being seperated headings, so I'm going to close this and I think I'll just change the heading to be `*NIX` or something similar as that list also includes cross-platform package managers like cargo or nix.

PR #839, opened 2021-10-26
Add support for smalltalk

jake-87:
Added support for the smalltalk programming language.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -1143,6 +1143,12 @@
       "multi_line_comments": [["(*", "*)"]],
       "extensions": ["sml"]
     },
+    "Smalltalk": {
+      "name": "Smalltalk",
+      "quotes": [["'", "'"]],
+      "multi_line_comments": [["\\\"", "\\\""]],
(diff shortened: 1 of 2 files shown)


jake-87:
That is definitely be a problem. Afaik, smalltalk programs almost always have two extensions on the end, such as .cs.st or .pck.st

jake-87:
Would it be better to add both of those options as potential file extensions? E.G. `"extensions": ["cs.st", "pck.st"]`

jake-87:
@spenserblack Should be fixed now

spenserblack:
@jake-87 cool! I'm not a collaborator, though, so you'll probably want to ping @XAMPPRocky instead so they can run the workflows 🙂

jake-87:
Ah, sorry about that. @XAMPPRocky, workflows should be good to go

jake-87:
@XAMPPRocky ?

XAMPPRocky:

PR #847, opened 2021-11-18
update readme: add tokei-pie in related tools.

laixintao:
Hi, I wrote a small tool that can parse `tokei -o json`, use like this `tokei -o json | tokei-pie` then generate a sunburst chart, displaying it by opening a tab in the browser. It is interactive, you can click around the sectors to see the code lines reactively.

It parses `tokei`'s json output so this feature do not touch tokei's code at all.

I think it is handy working with tokei. So I think it may interest other tokei users, so I open this PR.

I know it may be rude for opening a PR to just promote myself's project like this, I totally understand if you don't want to accept this, you can just close it if so. :D

https://github.com/laixintao/tokei-pie

### DIFF
--- a/README.md
+++ b/README.md
@@ -530,6 +530,11 @@ The canonical source of this repo is hosted on
 [GitHub](https://github.com/XAMPPRocky/tokei). If you have a GitHub account,
 please make your issues, and pull requests there.
 
+## Related Tools
+
+- [tokei-pie](https://github.com/laixintao/tokei-pie): Render tokei's output to
+  interactive sunburst chart.
+
 ## Copyright and License
 (C) Copyright 2015 by XAMPPRocky and contributors
 


XAMPPRocky:
Thank you for your PR! This looks super cool, I can't wait to show it to other people. 😄

laixintao:
Thanks! it means a lot to me. Let me know if you have more ideas or feedbacks on this tool. :)

XAMPPRocky:
@laixintao The couple of things that I noticed, just trying it out.

- I initially foolishly ran `tokei | tokei-pie`, and I get a `JSONDecodeError` backtrace. It would be nice if the error was caught, and it printed a more informative error message like "Non valid JSON input, maybe missing `-o json` flag", or something like.
- I noticed it opens up a webserver to serve the result. It would be nice if I could output it to a file somewhere instead, and use `open`.

laixintao:
Thanks! Excellent advice.

1. json decode error will not print an error message then exist as code=128
2. you can now use `tokei -o json | tokei-pie -o a.html` to write chart to a *single* HTML file, copy it around or open it anytime, won't open browser when `-o` used.


image

PR #850, opened 2021-11-21
languages.json: Racket - add scrbl extension and racket env

xgqt:
Signed-off-by: Maciej Barć <xgqt@riseup.net>

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -1068,7 +1068,8 @@
       "line_comment": [";"],
       "multi_line_comments": [["#|", "|#"]],
       "nested": true,
-      "extensions": ["rkt"]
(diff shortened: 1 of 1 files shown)


xgqt:
Explaination:
- `scrbl` - Scribble is the official documentation language of Racket built on top of it, so it would make sense to include it while counting Racket source (it is really different than Markdown or Org since it, for example, supports declaring Racket functions), more info: https://docs.racket-lang.org/scribble/index.html
- `env` - Racket supports declaring a interpreter on top of the file `#!/usr/bin/env racket`.

XAMPPRocky:

Issue #771, opened 2021-05-19
Unexpected error message


zsynacl:
### Problem
run command like `tokei -t="C++,C++ Header"  -i ../rslt.json`, I got error message below:
```bash
Error:
 Failed to parse input file: ../rslt.json

This version of tokei was compiled without serialization support for the following formats:

    json

You may want to install any comma separated combination of "cbor,json,yaml":

    cargo install tokei --features "cbor,json,yaml"
Or use the 'all' feature:

    cargo install tokei --features all
```
### Reproduce steps
1. run `tokei -t="C++, C++ Header" -o=json > rslt.json
2. go to any child folder, like `src`
3. run `tokei -t="C++,C++ Header"  -i ../rslt.json` in child folder

### expected behavior
show result of child folder or show any exact error message rather than "json not supported"

### other information
```powershell
PS C:\Code> tokei --version
tokei 12.1.2 compiled with serialization support: json, cbor, yaml

PS C:\Code> rustup toolchain list
stable-x86_64-pc-windows-gnu (default)

PS C:\Code> echo $host
Name             : ConsoleHost
Version          : 5.1.19041.868
InstanceId       : (xxxxxxxxx)
UI               : System.Management.Automation.Internal.Host.InternalHostUserInterface
CurrentCulture   : zh-CN
CurrentUICulture : zh-CN
PrivateData      : Microsoft.PowerShell.ConsoleHost+ConsoleColorProxy
DebuggerEnabled  : True
IsRunspacePushed : False
Runspace         : System.Management.Automation.Runspaces.LocalRunspace
```

OS version:
Windows 10 Enterprise
Version: 2004
OS Build: 19041.868

LovecraftianHorror:
I've got an attempted fix for this issue in #861. Just to note the only commands needed to reproduce are outputting and inputting on the `tokei` repo

```console
$ tokei -o json | tokei -i stdin
```

PR #861, opened 2021-12-13
Fix input output mismatch

LovecraftianHorror:
Closes: #771 

This PR attempts to fix a couple of issues with the `--output` `--input` flags. The implementation is a breaking change from the old format as described below.

## First mismatch

Before serializing the output an extra `Total` count was added in while deserializing did not account for this and would try to parse `Total` as a `LanguageType`. This was one cause of the deserialization error

## Second mismatch

Serializing previously converted the top level of `LanguageType`s stored in the map to a `String` using `.to_string` before serializing (I'm assuming to get a `String` so that it could be stored with `"Total"`). This differs from how the `LanguageType` would normally get serialized using the `Debug` implementation causing a mismatch when trying to deserialize. E.g. `LanguageType::Bash` with `.to_string()` would be `BASH` while its debug representation is just `Bash`

## Attempted solution

The simplest solution seemed to be to separate out the totals from the full language listing. This avoids having to special case the extra field while providing a better structure for the outputted format.

A test case was added that ensures the serialized then deserialized listing matches the original one for all available formats.

---

This is my first time contributing to this project so let me know if anything looks off or can be improved :)

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -7,7 +7,9 @@ type LanguageMap = BTreeMap<LanguageType, Language>;
 
 #[derive(Deserialize, Serialize, Debug)]
 struct Output {
+    #[serde(flatten)]
     languages: LanguageMap,
+    #[serde(rename = "Total")]
     totals: Language,
 }
 
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,9 +1,15 @@
-use serde_json::{json, Map};
+use serde::{Deserialize, Serialize};
 use std::{collections::BTreeMap, error::Error, str::FromStr};
 
 use tokei::{Language, LanguageType, Languages};
 
 type LanguageMap = BTreeMap<LanguageType, Language>;
+
+#[derive(Deserialize, Serialize, Debug)]
+struct Output {
+    languages: LanguageMap,
+    totals: Language,
+}
 
 macro_rules! supported_formats {
     ($(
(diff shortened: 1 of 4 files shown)


src/input.rs
```diff
@@ -1,10 +1,16 @@
...
 use std::{collections::BTreeMap, error::Error, str::FromStr};
 
 use tokei::{Language, LanguageType, Languages};
 
 type LanguageMap = BTreeMap<LanguageType, Language>;
 
+#[derive(Deserialize, Serialize, Debug)]
+struct Output {
+    languages: LanguageMap,
+    totals: Language,
```
XAMPPRocky, line 11: To remove the breaking change aspect, I wonder if it could use `#[serde(flatten)]` and rename it back to "total" rather than "totals"?

```suggestion
    #[serde(flatten)]
    languages: LanguageMap,
    total: Language,
```

XAMPPRocky (commented):
Thank you for your PR! This mostly looks good to me, there's one small thing I'd like to try, but if it doesn't work that's fine.

src/input.rs
LovecraftianHorror: That's a good idea! It looks like `flatten` isn't an issue for this. Would you like me to do `#[serde(rename = "Total")]` to preserve the old capitalization for that bit as well?

And this would still be a breaking change since the names for the top level `LanguageType` fields may still be different (`Bash` vs `BASH`), but it keeps the old structure which likely would have been the more obtrusive part of the change

XAMPPRocky: Ah true, yeah keeping the same casing makes sense to me.

LovecraftianHorror: Sounds good. Changes are in 39f40b192337257f48288530d393610cd7232580

LovecraftianHorror:
That should fix up the one review comment. Let me know if there is anything else you want changed!

Issue #727, opened 2021-02-25
panic in trim_first_and_last_line_of_whitespace


mykter:
Given a Vue <script> with a carriage return (and no linefeed), followed by non-whitespace, followed by whitespace, we try and return a slice that ends before it starts.

It seems likely the logic around line 43 (which is looking for carriage returns specifically) is at fault.

```
$ echo -e '<script>\ra </script>'  > test.vue
$ RUST_BACKTRACE=1 ./target/debug/tokei ./test.vue
thread '<unnamed>' panicked at 'slice index starts at 2 but ends at 1', src/utils/ext.rs:55:10
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/core/src/panicking.rs:92:14
   2: core::slice::index::slice_index_order_fail
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/core/src/slice/index.rs:48:5
   3: <core::ops::range::Range<usize> as core::slice::index::SliceIndex<[T]>>::index
             at /home/mmacnair/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:236:13
   4: <core::ops::range::RangeInclusive<usize> as core::slice::index::SliceIndex<[T]>>::index
             at /home/mmacnair/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:404:9
   5: core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
             at /home/mmacnair/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:15:9
   6: <[u8] as tokei::utils::ext::SliceExt>::trim_first_and_last_line_of_whitespace
             at ./src/utils/ext.rs:55:10
   7: tokei::language::syntax::SyntaxCounter::parse_context
             at ./src/language/syntax.rs:471:25
   8: tokei::language::syntax::SyntaxCounter::perform_multi_line_analysis
             at ./src/language/syntax.rs:245:34
[...]

PR #865, opened 2021-12-18
Fix panic `trim_first_and_last_line_of_whitespace` on bad slice indices

LovecraftianHorror:
Fixes: #727 

I believe this appropriately fixes the panic found in #727 where the end index would end up preceding the start.

If this doesn't match the expected behavior of the method then let me know. I didn't see any comments or example tests, so I was mostly just guessing at how it should behave for that case. It seemed like the condition checking for `self[i.saturating_sub(1)] == b'\r'` was to handle stripping off the full `b"\r\n"`, so I updated the condition to handle that case specifically

The end index could also be out of the range of the slice too since it's fallback value is `self.len()` while the slicing uses an inclusive range

I ended up adding a basic property-based test along with regression tests to be more confident in the implementation which is what ended up catching the case with end index's fallback value


XAMPPRocky:
Thank you for your PR! I’ve been quite busy with other projects so I definitely appreciate someone coming along and fixing these issues. 🙂

LovecraftianHorror:
`tokei` is one of the very first rust projects I learned of way back when I first started learning Rust, so I'm just happy to be able to contribute back :)

PR #868, opened 2021-12-19
Add blobs to the totals for jupyter

### LINKED ISSUE #859: Jupyter Notebooks lines does not contribute to total sum
Hi!

It seams to me that `.ipynb` files get counted, but somehow do not contribute to the grand total.

Example with the table from below:
415 lines of JavaScript + 2536 lines of Python = 2951 lines. Note that we have not added the "Jupyter Notebooks" lines here. 2951 is the number in the last line of the table (grand total).
Correct would be 415 lines of JavaScript + 2536 lines of Python + 264 lines of Jupyter Notebooks = 3215 lines total.

```
❯ tokei src/bachelorarbeit simulation_results_and_analysis/*.ipynb                                  
===============================================================================
===============================================================================
 JavaScript              5          415          300           30           85
 Python                 35         2536         2142           68          326
 Jupyter Notebooks       4            0            0            0            0
 |- Markdown             2           30            0           28            2
 |- Python               4          234          222            7            5
 (Total)                            264          222           35            7
===============================================================================
 Total                  44         2951         2442           98          411
===============================================================================
```
Tokei version:
```
❯ tokei -V
tokei 12.1.2 compiled with serialization support: json, cbor, yaml
```
Counting results without Jupyter Notebook for comparison:
```
❯ tokei src/bachelorarbeit                                        
===============================================================================
===============================================================================
 JavaScript              5          415          300           30           85
 Python                 35         2536         2142           68          326
===============================================================================
 Total                  40         2951         2442           98          411
===============================================================================
```
I assume what happens is that the line `Jupyter Notebooks       4            0            0            0            0` gets added to the grand total. Because the number of files is correctly count in grand total. The other columns are not added to the grand total.

LovecraftianHorror:
Fixes #859 

It looks like the blobs when getting stats for jupyter notebooks were getting added as children, but not being added to the root count, so this change simply adds them to the root count as well

### DIFF
--- a/src/language/language_type.rs
+++ b/src/language/language_type.rs
@@ -260,7 +260,8 @@ impl LanguageType {
             .collect::<Vec<_>>();
 
         for (language, stats) in iter {
-            *jupyter_stats.blobs.entry(language).or_default() += stats;
+            *jupyter_stats.blobs.entry(language).or_default() += &stats;
+            jupyter_stats += &stats;
         }
 
         Some(jupyter_stats)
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR!

PR #869, opened 2021-12-29
Add basic Haml support

qnighy:
As seen in https://github.com/XAMPPRocky/tokei/issues/587 and https://github.com/XAMPPRocky/tokei/pull/597, Haml uses a unique syntax that is hard to integrate into tokei. Though it is better to have basic support than nothing.

Alternative proposal: https://github.com/XAMPPRocky/tokei/pull/870


XAMPPRocky:

PR #870, opened 2021-12-29
Add Haml support with indented comments and embeddings

qnighy:
This is an alternative proposal to https://github.com/XAMPPRocky/tokei/pull/869.

This PR adds support for Haml with indented comments and embeddings like:

```haml
  -# Embeddings
  :javascript
    // This is JavaScript code:
    console.log("Hello!");

  -# Comments
  /
    %p
      This is emitted in an HTML comment.
```

Compared to https://github.com/XAMPPRocky/tokei/pull/869 this is more feature-complete but it makes the Rust implementation more complex. That is why I filed it as a separate proposal.

Known problems:

- Trailing empty lines in embeddings are entirely dropped from counting.
- Trailing empty lines in indented comments are counted as comments, not blanks.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -532,6 +532,11 @@
       "quotes": [["\\\"", "\\\""]],
       "extensions": ["gw"]
     },
+    "Haml": {
+      "line_comment": ["-#"],
+      "quotes": [["\\\"", "\\\""], ["'", "'"]],
+      "extensions": ["haml"]
+    },
     "Happy": {
       "extensions": ["y", "ly"]
     },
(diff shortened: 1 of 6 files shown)


XAMPPRocky:
Thank you for your PR! Sorry for the delay I'd like to consider allowing indented comments, however, I'd like to know that it would be useful for more than a single language. Do you know of other languages that use indented comments that could use this?

XAMPPRocky:
I'm going to go with the simpler solution for now, but if you know of multiple languages that use indented comments, and fix the known problems you listed in the PR, I'd be happy to revisit this.

PR #887, opened 2022-02-19
Add env match for Groovy

johnkeeping:
The standard #!/usr/bin/env is documented for Groovy [1] so add the
match for this.

[1] https://groovy-lang.org/syntax.html#_shebang_line


XAMPPRocky:

Issue #892, opened 2022-03-08
Liquid files embed support


eloyesp:
Liquid files are most of the times `html` files, so they can embed javascript and stylesheets (that are supported on HTML, but not on liquid files.

But it would be great to also support some liquid specific embeds:

- The schema tag embeds a `JSON`:

``` liquid
{% schema %}
{
  "name": "Article",
  "class": "article-page"
}
{% endschema %}
```
- The `style` tag embeds `CSS`:
- 
``` liquid
{% style %}
.blog-article .article-container p {
  font-family: {{ section.settings.font_family }};
  margin: 0 auto 60px;
}
{% endstyle %}
```

And the other missing embed that is getting quite common, is missing also on HTML, that is the SVG tag:

``` HTML
<svg xmlns="http://www.w3.org/2000/svg" width="18" height="17" viewBox="0 0 18 17">
  <path fill="#000" d="M16.5 7.1l-1.4-1-1.5z"></path>
</svg>
```

Can I help adding support for that?

PR #893, opened 2022-03-09
Fix liquid embeds

eloyesp:
I'm just trying to check what can be implemented about #892 without changing a lot (as it is the first time I touch a rust project).


eloyesp:
The problems I've found so far are that most of the code to found the embeds is hardcoded, making it hard to add new embed types (comparing with adding a different syntax), It would be nice to add the required regexes to the `languajes.json` file, instead of hardcoding those (that would lead to some repetition on the json file, but would make it easier to add new embed syntaxes).

PR #914, opened 2022-05-27
add Puppet to tokei; steal .pp from Pascal

Roguelazer:
Add support for counting lines in [Puppet manifests](https://puppet.com).

Note that some Pascal files do use the `.pp` file extension (although I think `.pas` is more common in the wild); I suspect that Puppet is probably more common than Pascal, so it's probably a fair tradeoff to drop the second-most-common pascal extension.


XAMPPRocky:
Thank you for your PR!

Issue #915, opened 2022-06-03
backslash bug


sceneq:
9 lines 6 code 0 comments 3 blanks expected
9 lines 4 code 0 comments 5 blanks actual

```java
public class A { public void run() {
"\\"; // 1 code + 2 blanks (count as code?)


"\\"; // 1 code + 3 blanks



} }
```

LovecraftianHorror:
Since I was confused about the issue, the expected and actual are flipped. Also worth noting that this is reproducible on `master` still, and the issue appears to be that all lines between `"\\"` will be counted as code since that wasn't really clear to me

XAMPPRocky:
Thank you for your issue! I'm not quite sure exactly why though I would have guess that the issue is something to do with the the quotes being incorrectly interpreted. I won't have to investigate further for awhile, but if someone wants to take this on, I'm happy to accept a PR that fixes this and adds the above case to the tests.

PR #916, opened 2022-06-04
Fix backslash bug (#915)

sceneq:


XAMPPRocky:
Thank you for your PR! You also need add the check for is_verbatim, as escapes don't work on those strings. Also can you add the original example from the issue as a test.

sceneq:
Where should I add the test?
This problem(#915) is not limited to java, but also occurs in python code such as
```python
"\\" # 1
# com1
# com2
"\\" # 2
# com3
# com4
# com5
```

XAMPPRocky:
> Where should I add the test?

You can update the Java, or Python test. We don’t need to check it in every language, we just need at least one of them testing it.

sceneq:
Fixed the condition and updated the java test

XAMPPRocky:

Issue #316, opened 2019-03-15
List of languages to potentially add


codesections:
I recently noticed the similar project [loccount](https://gitlab.com/esr/loccount) and it looks like it supports several languages that we currently don't.  I'm dropping them here in case anyone is looking for good languages to add.  (If this list isn't useful, please feel free to close this issue).

- [ ] ABC
- [ ] Algol60
- [ ] Arc
- [ ] Asciidoc
- [ ] Asm
- [ ] Autotools
- [ ] Awk
- [ ] B
- [ ] BASIC
- [ ] BCPL
- [ ] BLISS
- [ ] Batchfile
- [ ] CLU
- [ ] CML
- [ ] Chapel
- [ ] ChucK
- [ ] Cobra
- [ ] Csh
- [ ] Dylan
- [ ] Eiffel
- [ ] Es6
- [ ] Expect
- [ ] Factor
- [ ] Fantom
- [ ] Frege
- [ ] Hy
- [ ] Icon
- [ ] Io
- [ ] J
- [ ] Lex
- [ ] Livescript
- [ ] Logo
- [ ] M4
- [ ] MATLAB
- [ ] ML
- [ ] MUMPS
- [ ] Mal
- [ ] Man
- [ ] Metafont
- [ ] Modula
- [ ] Modula2
- [ ] Modula3
- [ ] Nroff/troff
- [ ] Oberon
- [ ] Occam
- [ ] PHP
- [ ] PL/1
- [ ] POP-11
- [ ] PostScript
- [ ] PowerShell
- [ ] Rebol
- [ ] Rexx
- [ ] SETL
- [ ] SGML
- [ ] SNOBOL4
- [ ] Sather
- [ ] Sed
- [ ] Seed7
- [ ] Simula
- [ ] Skew
- [ ] Smalltalk
- [ ] Texinfo
- [ ] Turing
- [ ] VRML
- [ ] VisualBasic
- [ ] Waf
- [ ] WebAssembly
- [ ] Wish
- [ ] Yacc
- [ ] Yorick
- [ ] Zephir

yjhmelody:
I implement wasm and llvm, see #323.
And I found that Lua's `multi_line` is not correct.

Abhinickz:
Add Template Toolkit File (.tt extension).

masklinn:
Missing: [cython](https://cython.org) (.pyx / .pxd / .pxi), which is currently completely invisible to tokei.

huonw:
[...]

PR #923, opened 2022-06-17
add m4 language definition

Roguelazer:
This is necessarily kind of fuzzy because m4 supports _changing_ the comment character with the [changecom](https://www.gnu.org/software/m4/manual/html_node/Changecom.html) macro, and `dnl` isn't _really_ a comment (because parenthesis are still matched inside it), but it's a start.

This is on #316 so presumably other people are also interested in it.


XAMPPRocky:

PR #935, opened 2022-08-01
[issue_934] from_shebang tera function uses starts_with

### LINKED ISSUE #934: tokei doesn't indentify python 2.7 files
I was gathering the history of lines of code for some project, and some of the files started with `#!/usr/bin/env python2.7`, and tokei wouldn't count them at all. tokei -v says `Unknown environment: python2.7`

adam-tokarski:
Fixes #934

This PR allows to cover versioning of language executables, like it can happen for Python. Currently there is `python`, `python2` and `python3` in this language env. Could just add `python2.7`, but we should add new env item for each new Python version, like `python3.4`, `python3.10`, etc. 
This soon may be somewhat cumbersome, so there is a proposition just to match using `starts_with` which should cover such all things. Main drawback is that it could cover things like `pythonxxx`, but not sure if that's a very big problem.

---
It's merely my second touch of **rust** at all (not mentioning the tera), so if there would be some better way to handle that, I'd love to hear that.

### DIFF
--- a/src/language/language_type.tera.rs
+++ b/src/language/language_type.tera.rs
@@ -380,7 +380,13 @@ impl LanguageType {
                     match word {
                         {% for key, value in languages -%}
                             {%- if value.env -%}
-                                {%- for item in value.env  %}| "{{item}}" {% endfor %}=> Some({{key}}),
+                                {%- for item in value.env  %}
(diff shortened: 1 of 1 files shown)


XAMPPRocky:

PR #952, opened 2022-10-10
Upgrade num-format to 0.4.3

bcmyers:
I'm the author of `num-format`, one of `tokei`'s dependencies. Unfortunately I went a long time without maintaining the crate. But today I updated it. The public API hasn't changed, but I upgraded a number of `num-format`'s old dependencies, notably...

* arrayvec 0.4 -> 0.7.2
* itoa 0.4 -> 1.0.4

This PR updates your version of `num-format` to the latest release (`0.4.3`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -31,12 +31,9 @@ dependencies = [
 
 [[package]]
 name = "arrayvec"
-version = "0.4.12"
+version = "0.7.2"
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Thank you for your PR! However I think this got upgraded from another PR in the time since, so it's not relevant anymore. Thank you for maintaining `num-format`! <3

PR #961, opened 2022-12-18
Add Hy support

nagy:
A dialect of Lisp that's embedded in Python.
More info about the language:

https://hylang.org/


XAMPPRocky:
Thank you for your issue!

Issue #888, opened 2022-02-21
Support for Arch Linux PKGBUILD files


AlienJust:
Is it possible to add support for PKGBUILD files?
https://wiki.archlinux.org/title/PKGBUILD

XAMPPRocky:
Thank you for your issue! Yes, any sufficiently known programming language is able to be included. All that is required is making a PR using the [contributing guide](../blob/master/CONTRIBUTING.md).

PR #972, opened 2023-02-06
Add arch's PKGBUILD files

wanesty:
#888 
> Support for Arch Linux PKGBUILD files

I wanted it too so here it is.

"PacmanMakepkg" could be renamed either "Pacman" or "Makepkg" if you believe the name is too long.


XAMPPRocky:

PR #984, opened 2023-04-18
Add HiCAD to languages.json

petrisch:
Hi @XAMPPRocky

This is for a closed source language specific to a CAD program from [ISD Germany](https://www.isdgroup.com/).
The point in adding this is, I have to maintain a lot of legacy code, and this tool would help me alot.

`.mac` is very generic I know, but since no one else claimed it yet, it would at least serve my purpose 😄


petrisch:
Sorry I used an old version :-(

PR #985, opened 2023-04-18
Add HiCAD to languages.json

petrisch:
Hi @XAMPPRocky

This is for a closed source language specific to a CAD program from [ISD Germany](https://www.isdgroup.com/).
The point in adding this is, I have to maintain a lot of legacy code, and this tool would help me alot.

.mac is very generic I know, but since no one else claimed it yet, it would at least serve my purpose 😄

btw. I added a testfile but have no idea how to test with it, other than well using it, which works.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -613,6 +613,11 @@
       "blank": true,
       "extensions": ["hex"]
     },
+    "HiCad": {
+      "name": "HICAD",
+      "line_comment": ["REM", "rem"],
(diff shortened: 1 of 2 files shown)


petrisch:
Ah missing the most obvious one, what a classic 🤦 
So this works luckily and I guess the build fail has nothing to do with this PR.

XAMPPRocky:
Yeah, I need to figure out why android has stopped working for no reason.

PR #989, opened 2023-04-22
Use XDG conventions on macOS too

utkarshgupta137:
`~/Library/Application Support` is an utterly inconvenient location for macOS users. Most CLI tools follow the `XDG conventions` on macOS as well. Replace the `dirs` crate with the `etcetera` crate to use it on macOS. Config locations for other platforms are unchanged.

Disclaimer: I've recently become a maintainer of the etcetera crate. AFAIK, it is the only crate that allows using XDG on macOS, while also supporting Windows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -309,26 +309,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dirs"
-version = "4.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


utkarshgupta137:
@XAMPPRocky The PR is failing due to a recent commit. I tried removing it from the history, but that doesn't seem to have worked.

XAMPPRocky:

PR #992, opened 2023-04-26
Fix Razor support

meenzen:
- Adds missing `C#` comments
- Adds missing `C#` quotes
- Adds support for Razor Components (`.razor` files)

Blocks like this will now be counted correctly:

```razor
@{
    // foo
    string foo = "bar";

    /*
    * bar
    */
    string bar = "foo";
}
```

Razor docs: https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-7.0


XAMPPRocky:

PR #999, opened 2023-05-20
Add Nuget Configuration Support

Embers-of-the-Fire:
**Changes:**
- Add support for NuGet Configs, using `xml` parser but as a new language.
- Fix `CONTRIBUTING.md`'s Bug Report part.

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -953,6 +953,12 @@
       "quotes": [["\\\"", "\\\""], ["'", "'"]],
       "extensions": ["nqp"]
     },
+    "NuGetConfig": {
+      "name": "NuGet Config",
+      "multi_line_comments": [["<!--", "-->"]],
+      "quotes": [["\\\"", "\\\""], ["'", "'"]],
+      "filenames": ["nuget.config", "packages.config", "nugetdefaults.config"]
+    },
     "Nushell": {
       "line_comment": ["#"],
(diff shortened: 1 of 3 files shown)


Embers-of-the-Fire:
I've add `serde(alias)` into tera templates, and I hope this won't lead to other problems. So when parsing data, now the two types of name are both supported, and when serializing I only use the human friendly variant.

If it's not a good idea, ping me and I'll change my solution.

languages.json
```diff
@@ -368,6 +375,12 @@
       "quotes": [["\\\"", "\\\""], ["'", "'"]],
       "extensions": ["ebuild", "eclass"]
     },
+    "EdgeQL": {
+      "name": "EdgeQL",
+      "line_comment": ["#"],
+      "quotes": [["'", "'"], ["\\\"", "\\\""], ["$", "$"]],
+      "extensions": ["esdl"]
+    },
```
XAMPPRocky, line 383: This should be two different languages, `EdgeQL` and `EdgeQL Schema` schema files are `esdl` and EdgeQL query files are `edgeql`.

Embers-of-the-Fire: Sorry for that mistake, I haven't used EdgeDB before. I've fixed it now.

XAMPPRocky:

Issue #1000, opened 2023-05-29
Bitbake rules match all '*.conf' files.


sjjf:
I'm not sure what the correct way to handle *.conf files is, but it seems unlikely that assigning them all to the Bitbake language is the appropriate choice.

This rule was added with commit da0841dd66ebabc8798ab141703d1aa94cb6df56 from August 2022. I assume the fix would be something like:

```

**Comment by sjjf** [comment]
I'm happy to make a pull request with that change, but for an 8 character change that seems a bit over the top . . .

**Comment by XAMPPRocky** [comment]
Thank you for your issue! Seems like an appropriate step to take for now. Can I ask what language are you using in your project that uses the `conf` extension?

All PRs large and small are welcome and appreciated :)

**Comment by sjjf** [comment]
This is a puppet environment - it's got a whole bunch of files of various types, including templates for config files in at least ten different formats, and plain config files in another ten different formats.

I don't know if a generic "config file" language config would be viable, there are probably some reasonable assumptions you could make ('#' or ';' for line comments would probably work for the majority of what you might call generic *.conf files), but they'd be necessarily broad . . . But it's pretty obvious to me that *no* specific language should lay claim to the .conf extension - that way lies madness.

PR submitted, #1001 - I didn't see anything in the contributors doco about sign off or anything like that, if I missed something let me know and I'll fix it.

**Comment by sjjf** [comment]
Yeah, five minutes looking through `/etc` on my local Linux machine showed 95%+ using the '#' line comment format, as you'd expect . . . aside from a bunch of XML files masquerading as "*.conf" files (gconf and fontconfig being the culprits there).

I /think/ that would be something like:
```
diff --git a/languages.json b/languages.json
index 9627450..0614431 100644
--- a/languages.json
+++ b/languages.json
@@ -247,6 +247,11 @@
       "multi_line_comments": [["(*", "*)"]],
       "extensions": ["v"]
     },
+    "Config": {
+        "name": "Generic config file",
+        "line_comment": ["#"],
+        "extensions": ["conf"]
+    },
     "Cpp": {
       "name": "C++",
       "line_comment": ["//"],
```
Though that might be claiming more than it should, if the only option for deciding what language rule to match is the extension then that's probably the best it could do.

I just submitted a PR with that, #1002.

PR #1001, opened 2023-05-29
Remove 'conf' extension from the Bitbake config.

sjjf:
This really can't be claimed by a single language, unless it's an attempt at a "generic" config file format.

This should resolve issue #1000 (as reported by myself).


XAMPPRocky:
Thank you for your PR!

PR #1007, opened 2023-06-08
Update `--languages` Output Format

Embers-of-the-Fire:
### Update `--languages` Output Format

Now it seems like this:

```
──────────────────────────────────────────────────────────────────
| Language                           Extensions                  |
|────────────────────────────────────────────────────────────────|
| ABNF                               abnf                        |
| AWK                                awk                         |
| ABAP                               abap                        |
| ActionScript                       as                          |
| ...                                ...                         |
──────────────────────────────────────────────────────────────────
```

> `EmojiCode` is treated differently, since emojis are usually rendered as two characters, so I add a `\u+200b` to render it correctly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ src/test
 .tokeirc
 results.csv
 node_modules
+*.code-workspace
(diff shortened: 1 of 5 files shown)


XAMPPRocky:
Thank you for your PR!

PR #1013, opened 2023-06-26
Ignore format commits for `languages.json`

spenserblack:
This registers formatting commits on `languages.json` as revs to be
ignored. GitHub uses the file in the blame view, and these commits add
unnecessary noise that aren't relevant to finding actual changes (e.g.
when a language was added).

You can read more about this file [here](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).

******

There are a few other commits that add some noise, like b8f90c2e6de62885cadb9875806003854d051010,
but since that's not *purely* a formatting commit, it's not included.


XAMPPRocky:
Thank you for your PR!

PR #1021, opened 2023-07-25
Recognize GNUmakefile

shynur:
<https://www.gnu.org/software/make/manual/html_node/Makefile-Names.html>:

> By default, when `make` looks for the makefile, it tries the following names, in order: `GNUmakefile`, `makefile` and `Makefile`.


shynur:
@XAMPPRocky Please add `GNUmakefile`

XAMPPRocky:

PR #1028, opened 2023-09-23
Fix LD Script language data

rhysd:
This PR fixes the following points for linker scripts support:

- `.ld` file extension is very common ([e.g. Linux kernel](https://github.com/torvalds/linux/blob/master/arch/x86/boot/setup.ld)).
- `//` line comment is now allowed for linker scripts. [The specification](https://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_chapter/ld_3.html) only mentions `/*` and `*/` as comments. And I confirmed ld actually causes an error as follows.
  ```
  error: kernel.ld:4: unknown directive: //
  >>> // hoge
  >>> ^
  ```

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -838,7 +838,6 @@
     },
     "LinkerScript": {
       "name": "LD Script",
-      "line_comment": ["//"],
       "multi_line_comments": [["/*", "*/"]],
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR!

PR #1031, opened 2023-10-05
add shell completions generation

NickyMeuleman:
I heavily rely on shell completions for all CLI tools I use, it requires me to memorize less and prevents spelling mistakes.

This PR adds a way to generate shell completions.

Example of it running on my machine, using the zsh shell and using tab to see all possible completions, and sending the output for fish shell completions to a file:
https://asciinema.org/a/K0EIiTGadqN8N5MgeudE5GxDe

I originally wanted to generate completions in `build.rs` but I did not succeed because of issues described here: https://internals.rust-lang.org/t/problems-with-using-include-in-build-rs-for-rs-code/8447, even if I duplicated all dependencies into build-dependencies the compiler couldn't figure out the `tokei::` and `crate::` `use` statements in `cli.rs`

This implementation adds the ability to print the contents of a completion file to stdout.

I'm still not very comfortable with Rust lifetimes, and the compiler suggested to use `'static`, it worked, but I lack the knowhow on how to prevent using a static lifetime.

While this PR works, I'm sure it's not ideal and I opted for a PR instead of an issue since it's easier to see what I'm talking about this way.

I would like to rework this to make the buildtime completions generation work, and maybe even update the build pipeline in the future to include those in repo packages, but I'll need some help.

### DIFF
--- a/.github/workflows/mean_bean_ci.yml
+++ b/.github/workflows/mean_bean_ci.yml
@@ -64,7 +64,7 @@ jobs:
     needs: install-cross
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/download-artifact@v3
+      - uses: actions/download-artifact@v4
         with:
           name: cross-apple-darwin
           path: /usr/local/bin/
@@ -94,7 +94,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Download Cross
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
           name: cross-linux-musl
           path: /tmp/


Laura7089:
I'd love to see this merged!

src/main.rs
```diff
@@ -15,9 +19,21 @@ use crate::{
...
+fn print_completions<G: Generator>(gen: G, cmd: &mut Command) {
+    generate(gen, cmd, cmd.get_name().to_string(), &mut io::stdout());
+}
+
 fn main() -> Result<(), Box<dyn Error>> {
+let matches = build_cli().get_matches();
     let mut cli = Cli::from_args();
 
+    if let Some(generator) = matches.get_one::<Shell>("gencompletions") {
+        let mut cmd = build_cli();
```
fgimian, line 31: You should be able to do `Cli::command()` here instead of calling `build_cli()` 😄

fgimian:
I'm happy to also contribute to this. Just added a few minor comments after reviewing but it overall looks OK to me. @XAMPPRocky would you kindly let us know you would need to push this through?

Cheers
Fotis

PR #1040, opened 2023-11-20
Optimization: Read only first 128B from the file when searching for shebang

WieeRd:
Currently, `LanguageType::from_shebang()` uses `BufReader` and `BufRead::read_line()` to fetch the first line of the file.

- `BufReader::new()` allocates 8 KiB of buffer by default
- `BufRead::read_line()` can end up reading the entire file

A typical shebang line ranges from 10 to 30 characters (e.g. `#!/usr/bin/env python3` is only 22B),
and it is *very* unlikely the file contains a valid shebang syntax if we don't find a newline character within the first few dozens of bytes.

This PR modifies the code so that it only reads the first 128B of the file, into the fixed sized `u8` array, without involving the unnecessary `BufReader` and `String`.

### DIFF
--- a/src/language/language_type.rs
+++ b/src/language/language_type.rs
@@ -2,7 +2,7 @@ use std::{
     borrow::Cow,
     fmt,
     fs::File,
-    io::{self, BufRead, BufReader, Read},
+    io::{self, Read},
     path::{Path, PathBuf},
     str::FromStr,
 };
--- a/src/language/language_type.tera.rs
+++ b/src/language/language_type.tera.rs
@@ -356,16 +356,23 @@ impl LanguageType {
     /// assert_eq!(rust, Some(LanguageType::Rust));
     /// ```
     pub fn from_shebang<P: AsRef<Path>>(entry: P) -> Option<Self> {
-        let file = match File::open(entry) {
-            Ok(file) => file,
-            _ => return None,
-        };
+        // Read at max `READ_LIMIT` bytes from the given file.
+        // A typical shebang line has a length less than 32 characters;
+        // e.g. '#!/bin/bash' - 11B / `#!/usr/bin/env python3` - 22B
+        // It is *very* unlikely the file contains a valid shebang syntax
+        // if we don't find a newline character after searching the first 128B.
+        const READ_LIMIT: usize = 128;
 
-        let mut buf = BufReader::new(file);
-        let mut line = String::new();
-        let _ = buf.read_line(&mut line);
+        let mut file = File::open(entry).ok()?;
+        let mut buf = [0; READ_LIMIT];
 
-        let mut words = line.split_whitespace();
+        let len = file.read(&mut buf).ok()?;
+        let buf = &buf[..len];
+
+        let first_line = buf.split(|b| *b == b'\n').next()?;
+        let first_line = std::str::from_utf8(first_line).ok()?;
+
+        let mut words = first_line.split_whitespace();
         match words.next() {
             {# First match against any shebang paths, and then check if the
                language matches any found in the environment shebang path. #}


XAMPPRocky:
Thank you for your PR! Do you have a benchmark or similar that shows that this is faster than the current implementation?

WieeRd:
### *It was a misinput!*

WieeRd:
Little off-topic question, why does `from_path()` take unused `Config` parameter? Are there plans to utilize it somehow?

XAMPPRocky:
> Little off-topic question, why does from_path() take unused Config parameter? Are there plans to utilize it somehow?

I couldn't say why, I haven't focused on tokei in a few years.

WieeRd:
# Benchmarks

Are you still not convinced?

## Test Data

```bash
# ./chungus: 4 GiB of zeros
$ dd if=/dev/zero of=./chungus bs=4M count=1024
1024+0 records in
1024+0 records out
4294967296 bytes (4.3 GB, 4.0 GiB) copied, 5.50768 s, 780 MB/s

# ./script: tiny script with shebang
$ echo "#!/usr/bin/env python3\nprint('Hello World!')" > ./script
```

```text
$ ls
╭───┬─────────┬──────┬────────┬───────────────╮
│ # │  name   │ type │  size  │   modified    │
├───┼─────────┼──────┼────────┼───────────────┤
│ 0 │ chungus │ file │ 4.3 GB │ 2 minutes ago │
│ 1 │ script  │ file │   22 B │ 2 minutes ago │
╰───┴─────────┴──────┴────────┴───────────────╯
```

## `src/bin/bufreader.rs`

```rust
use std::{
    fs::File,
    io::{BufRead, BufReader},
    path::Path,
    time::Instant,
};

pub fn bufread(path: &Path) -> Option<usize> {
    let file = File::open(path).ok()?;

    let mut buf = BufReader::new(file);
    let mut line = String::new();

    buf.read_line(&mut line).ok()
}

fn main() {
    let t = Instant::now();
    bufread("./script".as_ref());
    let dt = t.elapsed();
    println!("./script: {}μs", dt.as_micros());

    let t = Instant::now();
    bufread("./chungus".as_ref());
    let dt = t.elapsed();
    println!("./chungus: {}μs", dt.as_micros());
}
```

## `src/bin/justreader.rs`

```rust
use std::{fs::File, io::Read, path::Path, time::Instant};

pub fn justread(path: &Path) -> Option<usize> {
    const READ_LIMIT: usize = 128;

    let mut file = File::open(path).ok()?;
    let mut buf = [0; READ_LIMIT];

    let len = file.read(&mut buf).ok()?;
    let buf = &buf[..len];

    let first_line = buf.split(|b| *b == b'\n').next()?;
    let first_line = std::str::from_utf8(first_line).ok()?;

    Some(first_line.len())
}

fn main() {
    let t = Instant::now();
    justread("./script".as_ref());
    let dt = t.elapsed();
    println!("./script: {}μs", dt.as_micros());

    let t = Instant::now();
    justread("./chungus".as_ref());
    let dt = t.elapsed();
    println!("./chungus: {}μs", dt.as_micros());
}
```

## Test Results

```bash
$ cargo run --release --bin justreader
    Finished release [optimized] target(s) in 0.01s
     Running `target/release/justreader`
./script: 5μs
./chungus: 7μs

$ cargo run --release --bin bufreader
    Finished release [optimized] target(s) in 0.02s
     Running `target/release/bufreader`
./script: 10μs
./chungus: 2619986μs
```

WieeRd:
> I haven't focused on tokei in a few years.

I started using tokei as a library in my recent projects, and hopefully I can find some time to improve the codebase and deal with unresolved issues in the span of next few months :)

WieeRd:
Rebased and force-pushed.

XAMPPRocky:

PR #1041, opened 2023-11-21
Add OpenQASM support

mtreinish:
This commit adds support for the OpenQASM language which is used in quantum computing. The specification for the language can be found here:

https://github.com/openqasm/openqasm

and is documented here:

https://openqasm.com/

and in the paper:

https://dl.acm.org/doi/10.1145/3505636


XAMPPRocky:

PR #1048, opened 2023-12-04
fixed language names not showing when in Light mode (light background

### LINKED ISSUE #805: Parts of the output are showing in white color on white background when running in WSL2 in Alacritty
Hi,

I recently switched to light mode & noticed that some parts of the output are white text on white background.
Everything works correctly in Alacritty when using Windows PowerShell - this only occurs when I switch to WSL2.
I checked my colors in alacritty config & everything is set correctly (meaning only "background" & "white" is set to #FFFFFF)

Here is how it looks:
example_a
Here with mouse-selected text:
example_b

Alexe1289:
Fixes #805
Changed the language names to be coloured in magenta, so it can work in Light mode (white background).
BEFORE
image

AFTER:
image

### DIFF
--- a/src/cli_utils.rs
+++ b/src/cli_utils.rs
@@ -250,7 +250,7 @@ impl<W: Write> Printer<W> {
             write!(
                 self.writer,
                 " {:<len$}",
-                name.bold(),
+                name.bold().magenta(),
                 len = lang_section_len
             )?;
         }


XAMPPRocky:
Thank you for your PR! any reason this is a draft?

Alexe1289:
@XAMPPRocky I pulled this out of the draft.

PR #1056, opened 2024-01-02
Add language support for RP2040 PIO ASM

willeccles:
Adds support for the PIO assembler language for the RP2040, including embedded C code. I have left the "detected" language from the embedded code snippets generic and hard-coded C for now, as it's possible they could add support for other languages (i.e., C++, Python) in the future. To accomplish this, I have included new patterns and language contexts and so on where necessary.

However, I did not know how to handle the fact that it's an error for a `% c-sdk {` to be unterminated by a matching `%}`. At the moment, it simply treats it the same as Markdown--unbalanced is just treated as a valid block. Feedback would be appreciated. Relevant code snippets: [here](https://github.com/willeccles/tokei/blob/22f5408e334c3d447bd9e902a655a6e6e128b104/src/language/syntax.rs#L525-L559) and [here](https://github.com/willeccles/tokei/blob/22f5408e334c3d447bd9e902a655a6e6e128b104/src/language/language_type.rs#L186-L191).

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,6 +467,7 @@ Perl
 Perl6
 Pest
 Php
+PioAsm
 Poke
 Polly
 Pony
--- a/languages.json
+++ b/languages.json
@@ -1062,6 +1062,13 @@
       "quotes": [["\\\"", "\\\""], ["'", "'"]],
       "extensions": ["php"]
     },
+    "PioAsm": {
+      "name": "RP2040 PIO ASM",
+      "line_comment": [";", "//"],
+      "multi_line_comments": [["/*", "*/"]],
+      "important_syntax": ["% c-sdk {"],
+      "extensions": ["pio"]
+    },
     "Poke": {
       "multi_line_comments": [["/*", "*/"]],
       "extensions": ["pk"]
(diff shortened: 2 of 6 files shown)


willeccles:
Reproduction for the issue: consider a simple PIO program with the following code:

```
.program something
nop

% c-sdk {
void func(void);
// <---- code block NOT terminated!
```

The assembler will reject this program, but tokei will treat it the same way it does Markdown code fences: if it's not terminated, the rest of the file is C, and there is only one comment added for the `% c-sdk {`. This is *technically* correct in this situation, despite the program being ill-formed.

Now let's slightly tweak this file. If you simply start another PIO program underneath this, the counting is now broken:

```
.program something
nop

% c-sdk {
void func(void);
// <---- code block NOT terminated!

.program another_program
nop
```

This is also not accepted by the assembler, and again, tokei considers the rest of the file to be C code. However, in this example, this is *not* correct--this counts PIO instructions as C code.

I am not sure how this should be handled. Any thoughts?

willeccles:
Just realized that Python is supported in passthrough sections as per datasheet section 3.3.10.2. I'll update to include that.

Issue #1053, opened 2023-12-22
How are the filenames listed by the '--files' option ordered?


Adamska1008:
I start a new project and use tokei to track code lines. It seems that the filenames are not sorted in lexicographical order with `--files` flag.  It confused me because all files in different directories are mixed together, here is what it looks like

```
$ tokei --files
===========================================================================================================================================================
 Language                                                                                        Files        Lines         Code     Comments       Blanks
===========================================================================================================================================================
 Go                                                                                                  8          332          273           12           47
-----------------------------------------------------------------------------------------------------------------------------------------------------------
 ./middleware/auth_middleware.go                                                                                  7            5            0            2
 ./main.go                                                                                                       64           54            0           10
 ./handlers/task_handlers.go                                                                                     23           12            0           11
 ./config/config.go                                                                                              48           40            1            7
 ./models/task_models_test.go                                                                                    36           33            0            3
 ./models/user_models.go                                                                                          9            7            0            2
 ./handlers/auth_handlers.go                                                                                     97           82            9            6
 ./models/task_models.go                                                                                         48           40            2            6
===========================================================================================================================================================
 Total                                                                                               8          332          273           12           47
===========================================================================================================================================================
```

I know the order of filenames is not really an important issue, but I think having order is always better than disorder. I would appreciate hearing the opinions of others, if you don't mind.

PR #1059, opened 2024-01-18
Fixed --files argument to sort alphabetically

Alexe1289:
Now if --files argument is found it will sort by default alphabetically the name of the files. If --sort argument is also provided (for example tokei --files --sort blanks), the sorting for files will be ignored and it will only sort by given argument.

Should fix https://github.com/XAMPPRocky/tokei/issues/1053

### DIFF
--- a/src/cli_utils.rs
+++ b/src/cli_utils.rs
@@ -312,7 +312,7 @@ impl<W: Write> Printer<W> {
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


Alexe1289:
@XAMPPRocky any feedback, please?

XAMPPRocky:
Thank you for your PR! Sorry it will be awhile before I'm able to review.

XAMPPRocky:
Thank you for your PR!

PR #1067, opened 2024-02-02
Transforming input before reporting

Shogoki:
Hey there, 

We had the need to sometimes apply some transformations on the content of code files before getting the statistics for them. Right now `tokei` was not really giving us the possibility to do that.

Therefor I did take a look at the code and thought this would be easy to add, without affecting any existing functionality. Would be cool if we could get that integrated into an upcoming version.

What i did is, I added another Config option, which allows to pass a function, which will transform the text content of a file before getting the statistics.

Would love to hear your thoughts, and happy to adapt the feature in a way that might suit the project better.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -56,10 +56,9 @@ pub struct Config {
     pub for_each_fn: Option<fn(LanguageType, Report)>,
     #[serde(skip)]
     /// function to transform the file content before analysis
-    pub transform_fn: Option<TransformFn>,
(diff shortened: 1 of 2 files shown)


Shogoki:
Hey @XAMPPRocky,

Any chance you can take it a look at this and we can get this merged?

XAMPPRocky:
Thank you for your PR! However I'm quite busy right now, so it will be quite awhile before I can review this feature addition.

PR #1083, opened 2024-04-17
languages.json: add jinja extension for Jinja2

akx:
It's also quite common for Jinja2 templates to be called *.jinja; see [GitHub Code Search](https://github.com/search?q=path%3A*.jinja&type=code).


XAMPPRocky:

akx:
@XAMPPRocky Thank you for including this! I've been using Tokei for a long while, it's great 😄

PR #1084, opened 2024-04-24
feat: add support for Expect and update Tickle

OscarCreator:
[Expect](https://core.tcl-lang.org/expect/index) is an extension ontop of Tcl.


OscarCreator:
Do I need to simplify the tests or is an update needed?

PR #956, opened 2022-10-29
Add language support for Slang

halli2:
Adds support for the shader language Slang.

PR #1089, opened 2024-05-03
Fix slang

SandaruKasa:
#956 added the support, but didn't include the comment with the counts into the count of the comments.

This could've been caught by CI if it was run on Pull Requests.


XAMPPRocky:

> This could've been caught by CI if it was run on Pull Requests.

It is, it's just that running new CI jobs on older PRs isn't possible without a push.

Issue #1092, opened 2024-05-06
The value of `hidden` in the config file is not respected


SVendittelli:
## Summary

The value of `hidden` is not read from the config file (neither `tokei.toml` nor `.tokeirc`). The `hidden` field is documented on [Struct tokei::Config](https://docs.rs/tokei/latest/tokei/struct.Config.html).

## Expected behaviour

Running `tokei` with the following `tokei.toml` should have the same effect as running `tokei --hidden`:

```toml
hidden = true
```

## Actual behaviour

The value of `hidden` is ignored, setting `hidden = true` does not change the output of `tokei`.

## Steps to reproduce

1. In an empty directory, create `tokei.toml` with the following contents:
    ```toml
    hidden = true
    ```
2. Create `.hidden.toml` with the following contents:
    ```toml
    test = "data"
    ````
3. Run `tokei`:
    ```shell
    ===============================================================================
     Language            Files        Lines         Code     Comments       Blanks
    ===============================================================================
     TOML                    1            1            1            0            0
    ===============================================================================
     Total                   1            1            1            0            0
    ===============================================================================
    ```
4. Run `tokei --hidden`:
    ```shell
    ===============================================================================
     Language            Files        Lines         Code     Comments       Blanks
    ===============================================================================
     TOML                    2            2            2            0            0
    ===============================================================================
     Total                   2            2            2            0            0
    ===============================================================================
    ```

PR #1093, opened 2024-05-06
fix: read hidden from config file

SVendittelli:
Previously, setting `hidden = true` in the config file would have no effect. Now it is properly read.

Fixes: #1092


SVendittelli:
Rebased from master so the tests now pass

XAMPPRocky:

PR #1101, opened 2024-05-17
add Salesforce Apex language support

xixiaofinland:
Apex is a dialect diverged from JAVA, it has very similar syntx as Java.
It uses `.cls` and `.trigger` as file suffix.

However, I see VB6, a already supported language, also has `.cls`, not sure if this introduces conflicts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,6 +330,7 @@ Ada
 Agda
 Alex
 Alloy
+Apex
 Asn1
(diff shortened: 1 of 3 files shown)


xixiaofinland:
It seems the auto check indeed fails on `cls` file, please suggest how to proceed.

PR #1102, opened 2024-05-22
Add Spade

TheZoq2:
Adds support for https://spade-lang.org/


languages.json
```diff
@@ -1432,6 +1432,14 @@
       "quotes": [["\\\"", "\\\""]],
       "extensions": ["sol"]
     },
+
+    "Spade": {
+      "line_comment": ["//"],
+      "multi_line_comments": [["/*", "*/"]],
+      "nested": true,
+      "important_syntax": ["///"],
```
XAMPPRocky, line 1440: Why is this placed in the important syntax key?

TheZoq2: Since the syntax is very similar to Rust and it has that there. Both languages use `///` as doc comments, but maybe I misunderstood what that does

TheZoq2:
The CI failures look like they are not my fault, but let me know if you need me to do anything about it

PR #1121, opened 2024-07-03
Add hledger support

leana8959:
Hello,

I added the support for hledger (plain text accounting).

I noticed there are some differences between the comment syntax between "hledger" and "ledger", so I limited the extension to ".hledger", which should be hledger only.

hledger: https://hledger.org/1.34/hledger.html?highlight=extension#comments
ledger: https://ledger-cli.org/doc/ledger3.html#Transactions-and-Comments

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,6 +416,7 @@ Haskell
 Haxe
 Hcl
 Hex
+hledger
 Hlsl
 HolyC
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Thank you for your PR!

Issue #1141, opened 2024-08-20
Build warnings


qtfkwk:
```
warning: method `is_not_line_ending_whitespace` is never used
 --> src/utils/ext.rs:5:8
  |
3 | pub(crate) trait AsciiExt {
  |                  -------- method in this trait
4 |     fn is_whitespace(&self) -> bool;
5 |     fn is_not_line_ending_whitespace(&self) -> bool;
  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

warning: constant `FALLBACK_ROW_LEN` is never used
 --> src/consts.rs:3:11
  |
3 | pub const FALLBACK_ROW_LEN: usize = 81;
  |           ^^^^^^^^^^^^^^^^

warning: constant `LANGUAGE_COLUMN_WIDTH` is never used
 --> src/consts.rs:6:11
  |
6 | pub const LANGUAGE_COLUMN_WIDTH: usize = 10;
  |           ^^^^^^^^^^^^^^^^^^^^^

warning: constant `PATH_COLUMN_WIDTH` is never used
 --> src/consts.rs:7:11
  |
7 | pub const PATH_COLUMN_WIDTH: usize = 80;
  |           ^^^^^^^^^^^^^^^^^

warning: constant `FILES_COLUMN_WIDTH` is never used
 --> src/consts.rs:8:11
  |
8 | pub const FILES_COLUMN_WIDTH: usize = 8;
  |           ^^^^^^^^^^^^^^^^^^

warning: `tokei` (lib) generated 5 warnings
warning: field `verbose` is never read
  --> src/cli.rs:55:9
   |
38 | pub struct Cli {
   |            --- field in this struct
...
55 |     pub verbose: u64,
   |         ^^^^^^^
   |
   = note: `Cli` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
   = note: `#[warn(dead_code)]` on by default

warning: associated function `all` is never used
   --> src/cli_utils.rs:114:12
    |
104 | impl NumberFormatStyle {
    | ---------------------- associated function in this implementation
...
114 |     pub fn all() -> &'static [&'static str] {
    |            ^^^

warning: `tokei` (bin "tokei") generated 2 warnings
```

_Originally posted by @0323pin in https://github.com/XAMPPRocky/tokei/issues/1138#issuecomment-2299339915_

qtfkwk:
[...]

Issue #1138, opened 2024-08-20
Release v13.0.0-alpha.2 broken


9SMTM6:
If I run `cargo install --git https://github.com/XAMPPRocky/tokei.git tokei` then this builds successfully, but when I then run tokei I get on amd64 linux (arch linux based):

```
thread 'main' panicked at /home/[user]/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.5.15/src/parser/matches/arg_matches.rs:181:17:
arg `files`'s `ArgAction` should be one of `SetTrue`, `SetFalse` which should provide a default
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
zsh: IOT instruction (core dumped)  tokei
```

That error doesn't look platform specific.

When I specify the last alpha release it works: `cargo install --git https://github.com/XAMPPRocky/tokei.git --tag v13.0.0-alpha.1 tokei`.

XAMPPRocky:
Thank you for your issue! cc @qtfkwk it seems this is caused by your clap changes.

qtfkwk:
Can confirm @9SMTM6's error on a similar system (amd64 / debian). Looking into a fix now.

qtfkwk:
Found the same issue also affects the `hidden`, `languages`, `no_ignore`, `no_ignore_parent`, `no_ignore_dot`, and `no_ignore_vcs` options, since they are also flag options.

Have a working fix now. Submitting a PR momentarily.

qtfkwk:
Sorry about that! Please pull after the PR merges (or use my fork) and confirm if the issue is resolved. Thanks!

qtfkwk:
https://github.com/XAMPPRocky/tokei/pull/1139

0323pin:
oh ... I was coming here to say this ... I get the same.

But, I'll leave the build warnings as well
```
warning: method `is_not_line_ending_whitespace` is never used
 --> src/utils/ext.rs:5:8
  |
3 | pub(crate) trait AsciiExt {
  |                  -------- method in this trait
4 |     fn is_whitespace(&self) -> bool;
5 |     fn is_not_line_ending_whitespace(&self) -> bool;
  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

warning: constant `FALLBACK_ROW_LEN` is never used
 --> src/consts.rs:3:11
  |
[...]

PR #1137, opened 2024-08-17
Fix cargo audit issues

qtfkwk:
Related to: #901

Please review individual commits...

tldr:

* Ran `cargo outdated; cargo audit`
* Updated dependencies via `cargo upgrade -i && cargo update`
* Fixed issues preventing build (mostly due to clap v3 => v4)
* Fixed issues from `cargo clippy -- -D clippy::all`
* Cleaned up

PR #1142, opened 2024-08-20
Fix issue #1141

qtfkwk:
Summary of fixes:

* ~~Added *empty* `json` feature and included in `all`~~
* Removed unnecessary `Cli.verbose` property
* Removed unnecessary `NumberFormatStyle.all()` method
* Added doc strings for pub constants in `src/consts.rs`
* Exported `self::consts::*` in `src/lib.rs`
* Removed unused `AsciiExt.is_not_line_ending_whitespace()` method

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ version = "13.0.0-alpha.2"
 all = ["cbor", "yaml"]
 cbor = ["hex", "serde_cbor"]
 default = []
-json = []
 yaml = ["serde_yaml"]
 
 [profile.release]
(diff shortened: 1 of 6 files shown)


qtfkwk:
> * Added _empty_ `json` feature and included in `default`

Had to roll back the first item due to that fix causing build issues in checks.

AFAICT prior recent releases `v13.0.0-alpha{1,2}` also produced these warnings so it is not related to #1137 or #1138 or subsequent PRs.

Unsure if/how @XAMPPRocky et al want to address those?

XAMPPRocky:
Thank you for your PR!

PR #1143, opened 2024-08-21
Update README.md with HiCAD from d4a1814

petrisch:


XAMPPRocky:
Thank you for your PR!

Issue #1145, opened 2024-08-21
Release v13.0.0-alpha.3 broken


Sigmanificient:
- https://github.com/NixOS/nixpkgs/pull/336118

```
> results/tokei/bin/tokei --exclude nixpkgs
thread 'main' panicked at /build/tokei-13.0.0-alpha.3-vendor.tar.gz/clap_builder/src/parser/error.rs:32:9:
Mismatch between definition and access of `exclude`. Could not downcast to TypeId { t: (13369810092395217272, 7199936582794304877) }, need to downcast to TypeId { t: (12860632812102975081, 3350365521333402252) }

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted (core dumped)

> results/tokei/bin/tokei logs
thread 'main' panicked at /build/tokei-13.0.0-alpha.3-vendor.tar.gz/clap_builder/src/parser/error.rs:32:9:
Mismatch between definition and access of `input`. Could not downcast to TypeId { t: (13369810092395217272, 7199936582794304877) }, need to downcast to TypeId { t: (12860632812102975081, 3350365521333402252) }

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted (core dumped)

```

XAMPPRocky:
Cc @qtfkwk seems to be more clap related issues.

qtfkwk:
Confirmed behavior via:

```bash
~/github.com/qtfkwk/tokei$ git log -n 1 --pretty=format:"%h" --no-show-signature
612981e
~/github.com/qtfkwk/tokei$ cargo build
~/github.com/qtfkwk/tokei$ ./target/debug/tokei -e nixpkgs
thread 'main' panicked at src/cli.rs:327:50:
Mismatch between definition and access of `exclude`. Could not downcast to &str, need to downcast to alloc::string::String
```

qtfkwk:
PR: https://github.com/XAMPPRocky/tokei/pull/1146

qtfkwk:
Sorry about that @Sigmanificient! Thanks for your issue. Should be resolved following the PR and subsequent alpha release. Please confirm from your view.

rich-purnell:
When I was using version 13.0.0-alpha.4, this error still occurred.

```
╭─u0_a288@localhost ~/Source/SJTU-IPADS/OS-Course-Lab ‹main›
╰─$ tokei --version                                                                                                                                               134 ↵
[...]

PR #1146, opened 2024-08-22
Fix issue #1145

qtfkwk:
https://github.com/XAMPPRocky/tokei/issues/1145


qtfkwk:
Ref: https://docs.rs/clap/latest/clap/parser/struct.ValuesRef.html#examples

PR #1151, opened 2024-08-25
Add BQN support

slotThe:
BQN is an array language in the spirit of APL/J/K; [website here](https://mlochbaum.github.io/BQN/)


XAMPPRocky:
Thank you for your PR!

PR #1152, opened 2024-08-25
Add APL support

slotThe:
It's—quite literally—"a programming language" :)

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -33,6 +33,12 @@
       "multi_line_comments": [["/*", "*/"]],
       "extensions": ["als"]
     },
+    "Apl": {
+      "name": "APL",
+      "line_comment": ["⍝"],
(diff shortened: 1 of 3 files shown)


languages.json
```diff
@@ -33,6 +33,12 @@
       "multi_line_comments": [["/*", "*/"]],
       "extensions": ["als"]
     },
+    "Apl": {
+      "name": "APL",
+      "line_comment": ["⍝"],
+      "extensions": ["apl", "aplf", "apls"],
```
slotThe, line 39: APL is very old, and thus implementation are quite scattered—not to mention many are proprietary. Even worse, people aren't very keen on storing APL inside of plain text files anyways (though that thankfully appears to change a bit), meaning I've possibly missed a few extensions. These seem like quite save bets though.

XAMPPRocky:

PR #1160, opened 2024-09-07
Fix `.libjsonnet` extension for Jsonnet

eduardosm:


eduardosm:
I've actually got it wrong

PR #1175, opened 2024-09-28
Add support for justfiles

### LINKED ISSUE #1116: [Feature Request]: justfile
[just](https://github.com/casey/just) is a command runner that is kinda similar to `make` or `cmake`, it uses `justfile` to store its script. 

It would be great if it can be supported by `tokei`.

I am willing to submit a PR for this if the maintainer also wants this.

marcaddeo:
This adds support for [justfile](https://github.com/casey/just).

Just supports recipes that contain a shebang, but unfortunately it's being interpreted as a comment line and I don't see a way to resolve it within Tokei.

```just
# Comment
do-something:
  #!/usr/bin/env node
  console.log('im running in node')

# Comment
do-something-else:
  echo "regular recipe"
```

Closes #1116

### DIFF
--- a/README.md
+++ b/README.md
@@ -442,6 +442,7 @@ Json
 Jsx
 Julia
 Julius
+Just
 KakouneScript
 Kotlin
 Lean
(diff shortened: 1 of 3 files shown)


XAMPPRocky:

PR #1184, opened 2024-10-30
Add missing extension `fsti` for F*

W95Psp:
This PR just adds `fsti` to the list of extensions for F*. (while `fst` files are for implementations, `fsti` files are for interfaces)


XAMPPRocky:
Thank you for your PR!

PR #1192, opened 2024-11-15
Add support for 8th language

ErikSchierboom:
This PR adds support for the [8th language](https://8th-dev.com/)

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -456,6 +456,14 @@
       "line_comment": [";"],
       "extensions": ["edn"]
     },
+    "Eight": {
+      "name": "8th",
+      "line_comment": ["\\\\ ", "-- "],
+      "multi_line_comments": [["(*", "*)"]],
+      "nested": true,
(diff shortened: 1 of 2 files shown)


languages.json
```diff
@@ -1,5 +1,13 @@
 {
   "languages": {
+    "_8th": {
```
XAMPPRocky, line 3: I think it would be better to rename this, and move it into the E section
```suggestion
    "Eighth": {
```

ErikSchierboom: Good idea!

languages.json
```diff
@@ -456,6 +456,14 @@
       "line_comment": [";"],
       "extensions": ["edn"]
     },
+    "Eight": {
```
XAMPPRocky, line 459:
```suggestion
    "Eighth": {
```

XAMPPRocky:
Thank you for your PR!

PR #1201, opened 2024-11-25
Support for Eiffel programming language

elviejo79:
Eiffel is a very important language in the history of OOP. [Wikipedia](https://en.wikipedia.org/wiki/Eiffel_(programming_language))

Comments start with '--'
but comment for methods what we would consider doc strings are '--|'

### DIFF
--- a/tests/data/asn1.asn1
+++ b/tests/data/asn1.asn1
@@ -1,34 +1,25 @@
--- 34 lines 16 code 11 comments 7 blanks
-PKCS-12 {
-    iso(1) member-body(2) us(840) rsadsi(113549) pkcs(1) pkcs-12(12)
-        modules(0) pkcs-12(1) }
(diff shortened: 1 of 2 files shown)


tests/data/asn1.asn1
```diff
@@ -1,34 +1,25 @@
--- 34 lines 16 code 11 comments 7 blanks
-PKCS-12 {
-    iso(1) member-body(2) us(840) rsadsi(113549) pkcs(1) pkcs-12(12)
-        modules(0) pkcs-12(1) }
+-- 25 lines 16 code 3 comments 6 blanks
```
XAMPPRocky, line 1: This changes a file instead of adding a new one.

elviejo79:
@XAMPPRocky could you approve this pull request?
I think I have corrected everything.

PR #1214, opened 2024-12-15
Remove trailing commas from languages.json

C3pa:
fix: remove trailing commas from languages.json


XAMPPRocky:
Thank you for your PR, but these are intentional, the file is a JSON5 file which allows trailing commas.

C3pa:
It's fine then. I mostly thought keeping it consistent with most other entries was desirable.

PR #1237, opened 2025-02-20
fix panics from `clap` internal type casting

### LINKED ISSUE #1236: panics when passed `--num-format` flag
Parsing the `--num-format` flag causes a panic.

---

### Steps to reproduce
- \>= `tokei 13.0.0-alpha.8`
- pass the `--num-format` or `-n` flag

---
All 4 possible values for the flag cause the same result.

I was able to reproduce this using both the AUR `tokei-git` package on `tokei 13.0.0-alpha.8` and cloning the running the latest source from github. The stable `12.1.2` on nixpkgs did not have this issue.

Interestingly they had different panic messages.

Source code code from github (debug profile):
```
   ~/../../tokei ❯ ./target/debug/tokei ../../ninja -n dots
thread 'main' panicked at src/cli.rs:259:14:
Mismatch between definition and access of `num_format_style`. Could not downcast to tokei::cli_utils::NumberFormatStyle, need to downcast to alloc::string::String

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Release profile (both github src code and aur tokei-git package):
```
   ~/../../tokei ❯ tokei ../../ninja -n dots
thread 'main' panicked at /home/riley/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.5.15/src/parser/error.rs:32:9:
Mismatch between definition and access of `num_format_style`. Could not downcast to TypeId(0x87f994611b2542ce423a2635f1af69ff), need to downcast to TypeId(0xe66dcd7c1cbd9c1c240a693be4ce9ff4)

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
zsh: IOT instruction (core dumped)  tokei ../../ninja -n dots
```

Notably one places the panic in-crate and the other places the panic in an external crate.

---

### Relevant code snippet:
```rs
        // src/cli.rs:258-261
        let num_format_style: NumberFormatStyle = matches
            .get_one::<NumberFormatStyle>("num_format_style") // Panic line (according to debug build)
            .cloned()
            .unwrap_or_default();
```

---

I will bisect when I get a chance. Bad commit is on the version range `(12.1.2, 13.0.0-alpha.8)`

In the meantime, is anyone else able to reproduce?

JackDyre:
Resolves #1236

CLI flags that were constructed with the `.value_parser(["slice", "of", "possible", "values"])` would cause clap to panic due to a type casting error whenever parsed into any type other than `String`, (even `&str`).

The best solution is to parse to a `String` and map it to parse it further. The `value_parser!()` macro is more concise and allows clap to parse to the type specified, but you lose the error message that tells the user the possible values to pass to the flag.

Tests pass

A follow-up PR that adds test coverage to the CLI flags might be a good idea so this doesn't happen again.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -256,8 +256,8 @@ impl Cli {
         });
 
         let num_format_style: NumberFormatStyle = matches
-            .get_one::<NumberFormatStyle>("num_format_style")
-            .cloned()
+            .get_one::<String>("num_format_style")
+            .map(parse_or_exit)
             .unwrap_or_default();
 
         let number_format = match num_format_style.get_format() {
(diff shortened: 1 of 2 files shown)


JackDyre:
Oops I didn't notice until now that I did all of this just to be a duplicate of #1233.

That said I believe my PR is more robust as it handles panics in more than just the `--num-format` flag, and I believe the other PR has a logic error where it will not exit after being unable to parse the flag value.

JackDyre:
Oops I just noticed this PR is also a duplicate of #1209. Smh I need to do better investiagtion.


I still believe my implementation is better as per my comment about `value_parser!()` and fixing the other affect CLI flags

PR #1238, opened 2025-02-20
[CI] fixed broken CI due to deprecated action version

JackDyre:
- bump `upload-artifact` from v3 to v4 [[...](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/)]


JackDyre:
I had forgotten to bump `download-artifact` as well. It should be working now.

PR #1247, opened 2025-04-01
Add support for Veryl language

dalance:
This PR adds support for [Veryl](https://veryl-lang.org).


dalance:
I think #1287 is better than this PR, so please use it.

PR #1251, opened 2025-04-08
Add support for CUDA headers

hexclover:
* Add CUDA header language

* Add test data

* Update README


kjmeagher:
I use cuh files extensively, Is there a reason why this can't be merged?

PR #1253, opened 2025-04-14
Update etcetera to 0.9 and then to 0.10

musicinmybrain:
I confirmed that `cargo test --workspace` still passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -409,13 +409,13 @@ dependencies = [
 
 [[package]]
 name = "etcetera"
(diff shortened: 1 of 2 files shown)


musicinmybrain:
Now available is [`etcetera` 0.11](https://github.com/lunacookies/etcetera/releases/tag/v0.11.0), which has MSRV 1.87 and in exchange is able to drop its dependency on the `home` crate. Happy to follow up with a further update if there is interest.

PR #1254, opened 2025-04-25
Add support for modern Visual Studio Solution files (slnx)

karl-sjogren:
This PR adds support for the new XML-based solution format for Visual Studio, which is just XML with another extension.

For more info: https://devblogs.microsoft.com/dotnet/introducing-slnx-support-dotnet-cli/

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -1946,6 +1946,12 @@
       "blank": true,
       "extensions": ["sln"]
     },
+    "VisualStudioSolutionSlnx": {
+      "name": "Visual Studio Solution",
+      "multi_line_comments": [["<!--", "-->"]],
+      "quotes": [["\\\"", "\\\""], ["'", "'"]],
+      "extensions": ["slnx"]
+    },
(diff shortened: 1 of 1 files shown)


languages.json
```diff
@@ -1946,6 +1946,12 @@
       "blank": true,
       "extensions": ["sln"]
     },
+    "VisualStudioSolutionSlnx": {
+      "name": "Visual Studio Solution",
```
karl-sjogren, line 1950: I'm not sure what happens if you run this in a folder with both a `.sln` and `.slnx` file in the same directory structure. Should this one maybe have a name such as "Visual Studio Solution (slnx)"?

XAMPPRocky: Thank you for your PR! I believe it will just show two rows with the same name. I think having something to distinguish it like what you mentioned would be a good idea.

karl-sjogren: I have updated the name to "Visual Studio Solution (slnx)".

Issue #1113, opened 2024-06-15
tokei seems to be dramatically mis-estimating Go code


masklinn:
I was looking at https://go.dev/src/net/http/server.go in a local copy, using tokei 12.1.2 reports

    ===============================================================================
     Language            Files        Lines         Code     Comments       Blanks
    ===============================================================================
     Go                      1         3911         2906          764          241
    ===============================================================================
     Total                   1         3911         2906          764          241
    ===============================================================================

This seems credible, however while the extreme ratio of comments to code in the early part of the file does dry up some further down the file it never actually disappears. And indeed, using some basic `rg` filters for classification reveals a very different picture:

    > rg '^$' src/net/http/server.go | wc -l # blank lines
         387
    > rg '^\s*//' src/net/http/server.go | wc -l # comments
        1340
    > rg -v '^(\s*//|$)' src/net/http/server.go | wc -l # code
        2184

I'm sure tokei has a smarter morel than trivial regex, but the discrepancies seem really extreme, especially for a language which should be relatively easy to parse and classify. For what that's worth, `loc` returns the same things as the rg/wc values.

PR #1275, opened 2025-08-16
Fix wrong estimation of code in go

ashvwinn:
Tokei was detecting a "/*" in a multiline string "``" as a comment in go.
This makes the rest of the code as a comment too resulting in wrong code and comment count.

Adding multiline string quotes "``" in the 'languages.json' file fixes this.

Example:
```go
package main

import "fmt"

func main() {
    s := `/*`
    fmt.Println(s);
}
```

This resulted in the code count being only 4 lines and comments count being 2 even when there are no comments.

This is the output tokei gave before the change:
image

This is the output tokei gave after the change:
image

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -699,7 +699,7 @@
     "Go": {
       "line_comment": ["//"],
       "multi_line_comments": [["/*", "*/"]],
-      "quotes": [["\\\"", "\\\""]],
+      "quotes": [["\\\"", "\\\""], ["`", "`"]],
       "extensions": ["go"]
     },
     "Gohtml": {


ASH-WIN-10:
also fixes the issue #1113

tokei was showing more code count because it considered the '"' in "``" as the start of the string.

Example:
```go
package main

import "fmt"

func main() {
    s := `"`  // this would be considered as a the start of a string making the rest of the lines be considered as code even though there are some comments in it
    // comment
    //
    fmt.Println(s);
}
```

image

ashvwinn:
@XAMPPRocky sorry for tagging, but is the project dead?

PR #1291, opened 2025-11-07
Don't spell Bash in all caps

verhovsky:
For example see how its name is written in its [official documentation](https://www.gnu.org/software/bash/manual/bashref.html)

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -148,7 +148,6 @@
       "extensions": ["bal"]
     },
     "Bash": {
-      "name": "BASH",
       "shebangs": ["#!/bin/bash"],
       "line_comment": ["#"],
       "quotes": [["\\\"", "\\\""], ["'", "'"]],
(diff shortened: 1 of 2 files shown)


DeflateAwning:
While things often start off as acronyms, their style often morphs over time as they become regular words.

Here's a short list of similar ones that rightfully lost their all-caps status over time:

* Laser – Light Amplification by Stimulated Emission of Radiation, now everyone just says “laser.”
* Radar – Radio Detection and Ranging, but we just say “radar.”
* Scuba – Self-Contained Underwater Breathing Apparatus, now it’s a regular word.
* Sonar – Sound Navigation and Ranging, everyone just says “sonar.”
* Fish (the shell) - Friendly Interactive SHell. It's just "Fish" though.

My point is that "Bash" is now just a regular word, and is not an acronym.

verhovsky:
@XAMPPRocky can you merge this please?

PR #1293, opened 2025-11-15
Add macro file extension `fnlm` for Fennel

folliehiyuki:
Ref: https://fennel-lang.org/reference#import-macros-load-macros-from-a-separate-module


XAMPPRocky:
Thank you for your PR!

PR #1333, opened 2026-02-24
add .github/workflows/release_artifacts.yml for prebuild binaries in release page

stevenleeS0ht:
This PR is to solve the problem mentioned in this discussion <https://github.com/XAMPPRocky/tokei/discussions/1330>

@XAMPPRocky , may be it is a good idea to try to issue a new release for an experiment.

### DIFF
--- a/.github/workflows/release_artifacts.yml
+++ b/.github/workflows/release_artifacts.yml
@@ -1,0 +1,70 @@
+name: release_artifacts
+
+on:
+  release:
+    types: [published]
+
+env:
+  CARGO_INCREMENTAL: 0
+  CARGO_NET_GIT_FETCH_WITH_CLI: true
+  CARGO_NET_RETRY: 10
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! However we already have a ci for releasing artifacts and there are glaring issues with this code that lead me to believe it was not reviewed or tested before being opened so I'm going to close it. Code changes need to be tested and reviewed before being opened.

stevenleeS0ht:
@XAMPPRocky 

> However we already have a ci for releasing artifacts

Which script or workflow YAML is for publishing the prebuilt artifacts?

>  there are glaring issues with this code 

Can you give me some hint, may be I can fix it. 

It is a fact that after `13.0.0`, there is no longer prebuilt binaries in release page.

Odonno:
This is painful.

PR #1335, opened 2026-03-10
Add separate Luau language support

StayBlue:
Previously, files with the `.luau` extension were treated as `.lua` files. The Luau language has detached pretty far from its Lua beginnings, adding features such as a native type system. It also has string syntax that should be modeled directly for correct counting that isn't valid in Lua, such as string interpolation.

It also doesn't have one-to-one compatibility with Lua code, which is enumerated [here](https://luau.org/compatibility/).

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,6 +430,7 @@ Lisp
 LLVM
 Logtalk
 Lua
+Luau
 Lucius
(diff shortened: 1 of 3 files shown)


Migelo:

PR #1336, opened 2026-03-10
Update strum/strum_macros to 0.28.0

musicinmybrain:
This doesn’t affect MSRV, and `cargo test` still passes.


XAMPPRocky:
Thank you for your PR!

PR #1347, opened 2026-04-20
replace term_size with terminal_size

alexanderkjall:
This removes a dependency without adding a new one, as terminal_size was already in the dependency tree.

term_size seems like it might be an abandoned project.

Based on: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/tokei/debian/patches/use-terminal-size.diff?ref_type=heads


Migelo:

PR #1352, opened 2026-04-30
Add AscendC support in languages.json

adah1972:
It's like C++. https://www.hiascend.com/cann/ascend-c

### DIFF
--- a/languages.json
+++ b/languages.json
@@ -58,6 +58,13 @@
       "quotes": [["\\\"", "\\\""]],
       "extensions": ["art"]
     },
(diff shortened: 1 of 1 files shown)


adah1972:
Sorry that they do not seem to provide English documentation.

There is some sample code, like:

<https://gitcode.com/cann/cann-samples/blob/master/Samples/0_Introduction/vector_add/main.asc>

It is a bit like CUDA, basically C++ with special keywords (`__aicore__`) and kernel invocation support (`<<< … >>>`).

PR #1361, opened 2026-06-14
README: fix option documentation

Managor:


Managor:
Oh TheRootDaemon was ahead of me.

PR #1378, opened 2026-09-20
Add Teal support

lunadialsys:
This pull request adds a new language, [Teal](https://teal-language.org).
Teal is a dialect of Lua that adds incremental strong type-checking on top of the existing language, similar to TypeScript.

I've added Teal to the languages.json file and written a test script.


lunadialsys:
Note: there is an edge case interaction with multiline strings that contain the literal `--`. This is a line comment in Lua, and thus by extension Teal, but tokei does not handle this properly:
```lua
print([["This is a multiline string." ~ Coders Daily, 2026
-- This isn't a comment!
]]) -- BUG: ^ the above is counted as a comment if the ]] is on the same line as it
```
I tried to mess with the languages.json, but I could not find a way to make tokei behave in this case.

This is, though, a _very_ rare scenario, so I don't anticipate this coming up for the majority of users...

(I only found this after pushing the first commit, so I marked the PR as draft while I figured things out.)

